Issue #9, opened 2022-12-14
Move crate into rust-bitcoin


tcharding:
Hi,

Some things to consider for this to be brought over into rust-bitcoin. These are my observations only, I do not speak authoritatively for the rust-bitcoin project. Also, forgive me if these things have been mentioned/discussed already.

- Consider fleshing out the manifest i.e., make the crate ready for release on crates.io
- Add a README.md file (mention MSRV)
- Add a CHANGELOG.md file
- Copy the `rustfmt.toml` from `rust-bitcoin` and run `cargo +nightly fmt`
- Copy the `clippy.toml` file from `rust-bitcoin` to configure `clippy` to use the MSRV 

### Bonus Points

- Improve rustdoc form (use conventional sections, headings, tense etc.)
- Use same coding convention configurations as `rust-bitcoin` (included below) and fix all `clippy` warnings introduced by these
- Use `#[cfg_attr(docsrs, doc(cfg(feature = "rand)))]` anywhere that uses `#[cfg(feature = "rand")]` (adds a little badge to docs saying requires feature X, saves explicitly mentioning this in rustdoc)

```rust
// Coding conventions.
#![deny(non_upper_case_globals)]
#![deny(non_camel_case_types)]
#![deny(non_snake_case)]
#![deny(unused_mut)]
#![deny(missing_docs)]
```

(Please note I did not study the logic closely, these comments are all just surface things).

Thanks man!

yancyribbens:
@tcharding 

```
Use #[cfg_attr(docsrs, doc(cfg(feature = "rand)))] anywhere that uses #[cfg(feature = "rand")] (adds a little badge to docs saying requires feature X, saves explicitly mentioning this in rustdoc)
```

About your comment here, when I open the docs page with:

```
cargo doc --open --features="rand"
```

It doesn't appear any different with or without `#[cfg_attr(docsrs, doc(cfg(feature = "rand)))]` AFAIKT.  Am I missing something?

tcharding:
We need to set `docsrs` using the `RUSTDOCFLAGS` env var, here is an example - you could throw the same in the ci script in #10 if you wanted.
```

# Build the docs if told to (this only works with the nightly toolchain)
if [ "$DO_DOCS" = true ]; then
    RUSTDOCFLAGS="--cfg docsrs" cargo rustdoc --all-features -- -D rustdoc::broken-intra-doc-links -D warnings || exit 1
fi
```

[...]

PR #12, opened 2022-12-20
Add rustdoc build to CI script

yancyribbens:
Add rustdoc build to CI script: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/issues/9

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,9 @@
 #![deny(unused_mut)]
 #![deny(missing_docs)]
 
+// Experimental features we need.
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
 use std::cmp::Reverse;
 
 #[cfg(any(test, feature = "rand"))]
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -30,3 +30,8 @@ if [ "$DO_LINT" = true ]
 then
     cargo clippy --all-features --all-targets -- -D warnings
 fi
+
+# Build the docs if told to (this only works with the nightly toolchain)
+if [ "$DO_DOCS" = true ]; then
+    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly rustdoc --all-features -- -D rustdoc::broken-intra-doc-links -D warnings || exit 1
+fi


src/lib.rs
```diff
@@ -3,6 +3,7 @@
 //! a given target amount.
 //!
 
+#![feature(doc_cfg)]
```
tcharding, line 6: Would be better if this was conditionally enabled because if not one cannot run `cargo test` with stable toolchain.
```rust
#![cfg_attr(docsrs, feature(doc_cfg))]
```

tcharding (commented):
While testing this locally I think I worked out why we have `+nightly` in the rust-bitcoin ci script. Even though it is not useful for use in the github action it allows running the script with `DO_DOCS=true` locally _without_ having to set `RUSTUP_TOOLCHAIN=nightly`. Anyways, IMO the `test.sh` changes here are fine but the `lib.rs` one needs fixing.

src/lib.rs
yancyribbens: @tcharding good catch, thanks man.   I've added the conditional in so `cargo test` does not fail on `stable`.

yancyribbens: @tcharding it looks like `#![cfg_attr(docsrs, feature(doc_cfg))]` causes CI to fail.  I looked at [rust-bitcoin/bitcoin/contribt/test.sh](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/contrib/test.sh) and it's missing the step `cargo build` which is the step that now fails AFAICT.  I'm curious why is this missing from rust-bitcoin?  If `cargo build` is added to rust-bitcoin, will that CI also fail?

yancyribbens: nm it looks like I had place the `cfg_attr` in the wrong place, although I'm still curious why rust-bitcoin doesn't do a `cargo build` with all features.

tcharding (commented):
ACK 72af05a08a78ace7e9478110ca8d5b5bc1674fdd

tcharding (approved):

yancyribbens:
@Tibo-lg does this look ok to merge?  Thanks!

Tibo-lg (approved):

PR #14, opened 2022-12-22
Do not return when waste is zero because waste could be negative

yancyribbens:
To keep pace with changes in Bitcoin Core, this PR removes the condition that returns early if a result with a waste of zero is found.  The matching Bitcoin Core PR is [here](https://github.com/bitcoin/bitcoin/pull/25495).

The other reason why I want to implement this change is so that we can get an accurate benchmark comparison between this implementation and Bitcoin Core.  This change essentially makes every search an exhaustive search of the search space since there is no known optimal solution to halt on.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,7 @@ pub trait Utxo: Clone {
 }
 
 /// Select coins first using BnB algorithm similar to what is done in bitcoin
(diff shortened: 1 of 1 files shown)


Tibo-lg (approved):

Xekyo:
I just had a look at this change. I have the impression that the implementation here seems to optimize for the smallest sum of UTXO values at least meeting the target, not at the waste score.

Issue #2, opened 2020-04-14
Add Benchmarking


yancyribbens:
Add benchmarking using Explore bench-marking with [criterion](https://github.com/bheisler/criterion.rs) for example.

PR #15, opened 2022-12-22
Feature/benchmark

yancyribbens:
This PR closes https://github.com/p2pderivatives/rust-bitcoin-coin-selection/issues/2

The benchmark I added matches the benchmark in Bitcoin Core so an accurate comparison can be made.  The current results for this benchmark:

```
test benches::bench_select_coins_bnb ... bench:   1,234,158 ns/iter (+/- 523)`
```

I also ran the benchmark for Bitcoin Core to see how it stacks up:
```
|               ns/op |                op/s |    err% |     total | benchmark
|--------------------:|--------------------:|--------:|----------:|:----------
|        1,021,008.00 |              979.42 |    4.9% |      0.01 | `CoinSelection`
```

It seems that the current rust implementation is ~200 ns slower, although I can think of some optimizations that can bring the results closer :nerd_face: .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,11 @@
 #![deny(missing_docs)]
 
 // Experimental features we need.
+#![cfg_attr(bench, feature(test))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
+#[cfg(bench)]
+extern crate test;
 
 use std::cmp::Reverse;
 
@@ -171,7 +175,9 @@ mod tests {
     }
 
     impl Utxo for MinimalUtxo {
-        fn get_value(&self) -> u64 { self.value }
+        fn get_value(&self) -> u64 {
+            self.value
+        }
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -326,3 +329,55 @@ mod tests {
...
+
+        let mut utxo_pool = Vec::new();
+
+        for _i in 0..1000 {
+            let u = MinimalUtxo { value: 1000 * COIN };
+            utxo_pool.push(u);
+        }
+
+        let u = MinimalUtxo { value: 3 * COIN };
+        utxo_pool.push(u);
```
Tibo-lg, line 373:
```suggestion
        let u = MinimalUtxo { value: 1000 * COIN };
        let mut utxo_pool = vec![u; 1000];
        utxo_pool.push(MinimalUtxo { value: 3 * COIN });
```

src/lib.rs
```diff
@@ -326,3 +329,55 @@ mod tests {
...
+
+        bh.iter(|| {
+            let result =
+                select_coins_bnb(1003 * COIN, CHANGE_LOWER, &mut utxo_pool.clone()).unwrap();
+            assert_eq!(2, result.len());
+            assert_eq!(1000 * COIN, result[0].value);
+            assert_eq!(3 * COIN, result[1].value);
```
Tibo-lg, line 380: The asserts are slowing down the benchmark, IMHO the point is not to test here so I think it's better to remove them.

Tibo-lg (changes requested):
Small nits otherwise looks good thanks!

src/lib.rs
Tibo-lg: Never mind I checked again and it doesn't seem to have any impact actually.

yancyribbens:
Thanks @Tibo-lg, should be good now.

tcharding (approved):
ACK 15d8f5ef05ee03d9258cdf3fe2963a31276b0b7d

Tibo-lg:
@yancyribbens I think you need to rebase after it's good to merge

Issue #20, opened 2023-01-05
Missed Solution


yancyribbens:
Recently I've found an edge case where a solution should be found but is not.  For example, consider the following UTXO set:

`{1, 2, 2.5, 4}`

And, if we are looking for a target of 7, the solution ought to be: `{1, 2, 4}`, however, in my testing, no match is found, even though the combination is possible.

Here is the test that should pass but fails:

```
#[test]
fn find_solution_test() {
    let utxo_pool = vec![
    MinimalUtxo { value: ONE_BTC },
    MinimalUtxo { value: TWO_BTC },
    MinimalUtxo { value: 2 * ONE_BTC + 50000000 },
    MinimalUtxo { value: FOUR_BTC },
];
```

see branch: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/tree/bug/incorrect-solution

yancyribbens:
closed by https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/28

yancyribbens:
closed by https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

PR #21, opened 2023-01-08
Examine all possible combinations

yancyribbens:
Closes https://github.com/p2pderivatives/rust-bitcoin-coin-selection/issues/20

Try all possible combinations and record the best match.  This is of course very slow and needs to be improved before is ready to use on large size UTXO pools.  The number of possible combinations of a pool size of `n`would be:

The summation (sigma) for k from i..n of `C(n, k)= n!/[k!(n-k)!] `

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,8 +17,7 @@
 #[cfg(bench)]
 extern crate test;
 
-use std::cmp::Reverse;
-
+use combinations::Combinations;
 #[cfg(any(test, feature = "rand"))]
 use rand::{seq::SliceRandom, thread_rng};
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -85,70 +84,83 @@ pub fn select_coins_bnb<T: Utxo>(
...
+            }
         }
-
-        remainder -= utxo_pool[m].get_value();
-        curr_selection[m] = false;
     }
 
-    best_selection
+    best_set
 }
```
Xekyo, line 164: Better! AFAICT, this should test all possible input sets, however, I would expect this approach to become extremely slow starting somewhere in the range of 30–50 UTXOs simply due to actually generating the entire combination space.

The advantage of using a Branch and Bound algorithm is that it allows to exclude large parts of the exponentially-sized search tree when you recognize certain circumstances.

src/lib.rs
```diff
@@ -182,114 +194,131 @@ mod tests {
 
     #[test]
     fn find_solution_1_btc() {
-        let utxo_match = find_solution(ONE_BTC, COST_OF_CHANGE, &mut UTXO_POOL.clone()).unwrap();
-        let expected_bool_vec = vec![false, false, false, true];
-        assert_eq!(expected_bool_vec, utxo_match);
+        let utxo_match = select_coins_bnb(ONE_BTC, COST_OF_CHANGE, &mut UTXO_POOL.clone()).unwrap();
+        assert_eq!(1, utxo_match.len());
+        assert_eq!(ONE_BTC, utxo_match[0].get_value());
```
Xekyo, line 199: Perhaps write these tests as a test function that takes a `target` and the expected input set as inputs. Then instead of having 11 functions, create a `map<target, input_set>` with 11 entries.

Xekyo: Additionally, it seems to me that most of your tests are focused on exact matches of the target value. It would be good if you added 1) more tests that fall in the range between `lower_bound` and `upper_bound`, 2) a test that verifies that input sets matching `upper_bound` are accepted, 3) some tests with less round numbers, perhaps even multiple possible solutions, to test that you get the best, not the first solution, 4) some tests with bigger UTXO pools than 4 UTXOs.

src/lib.rs
```diff
@@ -372,8 +418,8 @@ mod benches {
             let result =
                 select_coins_bnb(1003 * COIN, CHANGE_LOWER, &mut utxo_pool.clone()).unwrap();
             assert_eq!(2, result.len());
-            assert_eq!(1000 * COIN, result[0].value);
-            assert_eq!(3 * COIN, result[1].value);
+            assert_eq!(1000 * COIN, result[1].value);
+            assert_eq!(3 * COIN, result[0].value);
```
Xekyo, line 422: Perhaps just check whether there is one coin of value 1'000 present, and one coin of value 3 present, rather than expecting them at specific positions.

This test made me realize that `combinations` deduplicates repeat values which is much better than my assumption above, but in the real world, UTXOs will often be similar in value but not match exactly. I'd be curious what sort of run time you would get in this bench if you had 30 UTXOs with distinct values instead.

Xekyo:
Ah, I just reread your comment on rust-bitcoin. Yeah, waste metric would seem higher priority than speeding up BnB. In fact, I would probably implement handling of fee rates and _effective values_ first to ensure that your input selections can actually pay for the transaction, get that to work with a simple algorithm like SRD, then implement waste metric as a fitness function to pick among multiple candidates, then move on to more complex solution approaches like BnB.

yancyribbens:
@Xekyo thanks for the feedback.  I've added  `effective values` in a simple algorithm (SRD) https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22.   I think it would make sense to address the remainder of the comments you made when I create the next PR with an updated BnB selection algorithm.

yancyribbens:
closing in favor of https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

PR #22, opened 2023-01-11
Feature/add srd selection

yancyribbens:
I have renamed and moved the random function to it's own module.  This feature is similar to [SRD](https://github.com/bitcoin/bitcoin/blob/4586ae2da1a4532dff459d6824b10bd329a5e824/src/wallet/coinselection.cpp#L168) in core.  Also, I've added `effective_value` which is used for the sum calculation instead of `value`.  Lastly I did some minor refactoring so that the unit sizes used in the test match core.


src/single_random_draw.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cfg(any(test, feature = "rand"))]
+use rand::{seq::SliceRandom, thread_rng};
+
+/// Randomly select coins for the given target by shuffling the utxo pool and
+/// taking UTXOs until the given target is reached, or returns None if the target
+/// cannot be reached with the given utxo pool.
+/// Requires compilation with the "rand" feature.
+#[cfg(any(test, feature = "rand"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
+pub fn select_coins_srd<T: Utxo>(target: u64, utxo_pool: &mut [T]) -> Option<Vec<T>> {
```
Xekyo, line 13: I figure it might also be a good time to talk about the `target`. You are aiming to make different attempts to come up with an input set. Some of these attempts will aim to create change and some will try to create a changeless transaction. Yet, the target should remain the same across all these attempts. That means that the target can only be composed of the parts that are the same for every coin selection attempt: the amount paid to recipients plus the fees for the recipient outputs plus the fees for the transaction header. After that we still add inputs, but due to us counting only their `effective_value`, they each already bring their own fees.—That leaves one thing unaccounted for: the change output. You may want to think about where the amount for the change and its fees should get added, and how you make sure it's present in algorithms that will create change while it is not present in algorithms that avoid change.

src/lib.rs
```diff
@@ -156,26 +132,33 @@ mod tests {
...
     }
 
     impl Utxo for MinimalUtxo {
         fn get_value(&self) -> u64 {
             self.value
         }
+
+        fn get_effective_value(&self) -> u64 {
+            self.value - self.fee
+        }
```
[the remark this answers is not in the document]
yancyribbens: @Xekyo I think it's important to note that this is the concrete definition provided in the test suite.  In order to run the tests, a concrete definition must be provided similar to how the code calling this crate would provide the definitions for the abstract UTXO [type](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/405451929568422f7df809e35d6ad8f36fccce90/src/lib.rs#L27).  In other words, since the details of how the UTXO is computed don't specifically concern coin_selection, then the definitions should be provided by the calling crate.  The crate that depends on this crate would provide the method body for `get_effective_value`.  Does that makes sense or do you think the abstract type should change?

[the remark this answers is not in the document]
Xekyo: I would orient myself by how rust-bitcoin defines its UTXO objects given that your crate is aiming to be called from there eventually. If the UTXOs in rust-bitcoin already track _effective\_value_ that would be great of course, but a call to a coin selection function should generally come from a context where a feerate is known and an expected input weight might be more accessible than you think.

Xekyo:
BTW, I added this repository to my watch list, so if you ask something here, I should also see.

src/lib.rs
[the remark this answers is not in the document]
Tibo-lg: > I would orient myself by how rust-bitcoin defines its UTXO objects given that your crate is aiming to be called from there eventually.

AFAIK/CT there is no UTXO object defined in rust-bitcoin and I don't think the goal is for rust-bitcoin to call this crate but rather for a user to be able to use it next to rust-bitcoin. 

Also the reason for defining a trait instead of a struct was mainly to avoid requiring a conversion when calling the coin selection or forcing users to have to use the defined struct in their code (which in many cases could be lacking some information that they need). I don't have a strong opinion on which way it should go but thought relevant to give the reasons for the design.

src/single_random_draw.rs
Tibo-lg, line 13: What do you think about taking the RNG as a parameter, similar to [this](https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/key.rs#L201)?

Xekyo: I'm not sure I grok the linked Rust, but an advantage of having the seed for randomness as an input is that it allows you to pass a static seed in tests to make them repeatable.

src/lib.rs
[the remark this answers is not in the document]
Xekyo: I see, thanks for the clarification. Let me try to rephrase the point I was attempting to make. The _effective value_ of UTXOs depends on the feerate that a transaction is being built at. This means that the value can only be calculated dynamically either by the caller before calling the coin selection function, or by the coin selection function itself. Since it's a bit of a specific thing that only coin selection needs, it seems to me that it could be handled on this side of the call.

So really, the caller must pass to coin selection one of three packages of information so it can make an informed decision. Coin selection needs either the 1) _effective\_value_ on UTXOs, 2) the intended _feerate_ plus _weight of each UTXO_, or 3) the intended _feerate_ plus the _type of UTXO_ plus knowledge of the _weight of standard input types_.

src/single_random_draw.rs
yancyribbens: I think it's a good idea if as @Xekyo mentioned that it's a way to make the tests deterministic.  Is there any other reason you ask?  In either case I'm happy to look at adding RNG as a parameter.

The `<R: rand::Rng + ?Sized>` part is probably unfamiliar when compared to other languages and hard to grok at first - however it's what Rust calls a trait bound - which simply tells the compiler that the type `R` must have the following traits (must have a known size at compile time and implement the method signatures for the abstract type rand::Rng).

src/lib.rs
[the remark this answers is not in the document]
yancyribbens: > Also the reason for defining a trait instead of a struct was mainly to avoid requiring a conversion when calling the coin selection or forcing users to have to use the defined struct in their code (which in many cases could be lacking some information that they need). 

Seems like a good reason to keep the interface abstract.

> This means that the value can only be calculated dynamically either by the caller before calling the coin selection function, or by the coin selection function itself.

Or the other option is it's calculated dynamically by a UTXO library that coin-selection depends on (which is neither the calling application nor the coin selection lib).

> Since it's a bit of a specific thing that only coin selection needs, it seems to me that it could be handled on this side of the call

I think since we need this to move forward and there isn't a UTXO library we can import (and we wouldn't want to require the calling application to know how to calculate this), it would make sense to include the UTXO weight calculation logic in this crate (for now).  However, it seems to me the coinselection algorithm doesn't need to know the specifics of how things like the `effective_value` are computed.  Instead, there are certain properties that must hold true, for example that the `effective_value >= target_value` (please correct me if I'm wrong here).  At any rate, I think a lot of this debate is if a UTXO library would ever be useful outside of coinselection, which would mean moving it to a separate library would be better (in the future).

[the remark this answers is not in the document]
yancyribbens: oops I meant `effective_value >= value`.

src/single_random_draw.rs
Tibo-lg: > I think it's a good idea if as @Xekyo mentioned that it's a way to make the tests to be deterministic.

Yes you can use [`StepRng`](https://docs.rs/rand/latest/rand/rngs/mock/struct.StepRng.html) for that :)!

> Is there any other reason you ask?

`thread_rng()` requires std so if someone doesn't want to use std it's nice to be able to give something else. You could also have two functions, one that takes the RNG and one that doesn't but requires an extra feature (e.g. `rand-std`). That's what I had for for [the Schnorr API of secp256k1](https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/schnorr.rs#L125). Might be an overkill for now though so I'd personally go with just taking the RNG as parameter.

src/lib.rs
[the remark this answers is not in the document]
Tibo-lg: > This means that the value can only be calculated dynamically either by the caller before calling the coin selection function, or by the coin selection function itself..

Personally as a user of the library I'd much prefer the second option, as I don't want to have to start thinking about what effective value is and how I should go to compute it.

>  However, it seems to me the coinselection algorithm doesn't need to know the specifics of how things like the effective_value are computed. Instead, there are certain properties that must hold true, for example that the effective_value >= target_value (please correct me if I'm wrong here). At any rate, I think a lot of this debate is if a UTXO library would ever be useful outside of coinselection, which would mean moving it to a separate library would be better (in the future).

I'm not sure a UTXO library would be useful, IMHO it makes sense to have the computation within this crate. The question then is what information we need the user to provide to be able to do the computation. Of course first would be the fee rate, then we need information about the input weight. I can think of two ways: 

1. an enum per input giving the type of output being spend. For P2PKH or P2WPKH the weight can be computed  easily, for P2SH, P2WSH or P2TR we'd need a parameter like `max_script_length` or `max_witness_len`. We could also add some well known types like 2-of-2 P2WSH (though I'd probably do that later).
2. take an output descriptor (from the [rust-miniscript crate](https://github.com/rust-bitcoin/rust-miniscript)) and just call [max_satisfaction_weight]((https://docs.rs/miniscript/latest/miniscript/descriptor/enum.Descriptor.html#method.max_satisfaction_weight) (and add the other input field weight).

The second option feels cleaner but requires a dependency. Maybe a combination of both with a feature flag for miniscript could be an option, but might be better to just pick one for now (if it'd be me I think I'd got the miniscript way but I don't have a very strong opinion).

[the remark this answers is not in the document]
yancyribbens: Thanks for all the feedback, I'm not sure a UTXO lib will be useful either, however for now it seems we all agree it's best to add the needed fee computations in this library.  I'll look at getting this added as a new commit to this PR.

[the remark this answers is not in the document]
yancyribbens: This commit is a bit hack and slash right now but I wanted to see what people thing: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/b149195a48234eb5872107fb32b4e48d88670488

I noticed this [commit](https://github.com/rust-bitcoin/rust-bitcoin/commit/6d51e9255b969e142d0fae6d4cf3fcdd8cfae0cc) that was recently added which does a lot of the heavy lifting to compute the input weight.  However this requires the downstream application to use the rust-bitcoin `TxIn`.  Also, using the `TxIn` doesn't seem to be enough to get the value of the input.  The `TxIn` only provides the `Outpoint` which is the transaction index [output](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/transaction.rs#L52) of the corresponding.  It doesn't seem ideal to require the downstream application to pass the value of the input but that's what I've done for now.  There's probably a better way to structure this but may require a change to the upstream lib.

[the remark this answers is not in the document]
Tibo-lg: I don't think that will work. IIUC the commit you linked to uses the witness size of the input to compute the weight, but when you are doing coin selection you haven't yet signed the input (since the tx is not built yet) so I don't think it really make sense, unless I'm misunderstanding something. Or you would need to require the user to pre-fill the witness with some fake data but that doesn't seem convenient to me.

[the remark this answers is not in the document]
yancyribbens: > the commit you linked to uses the witness size of the input to compute the weight

There are two possible weights for the inputs, either `segwit` or `non-segwit`:

```
pub fn calculate_fee(input: &TxIn, is_segwit: bool, feerate: u64) -> u64 {
   let weight = if is_segwit {
        TxIn::segwit_weight(input)
    } else {
        TxIn::legacy_weight(input)
    };

    weight as u64 * feerate
}
```
 
> but when you are doing coin selection you haven't yet signed the input (since the tx is not built yet) so I don't think it really make sense

I'm not sure about that right now.  My first thought is that I would think the signature is a known size ahead of time depending on if the transaction is segwit or not.  I guess my thinking is that we are only looking to examine the weight of each input to determine the feerate and subtract that feerate from the input value (effective_value).  I'm not sure how signing one or more inputs impacts the calculation.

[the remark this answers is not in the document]
Tibo-lg: The problem is that you don't know what data is required to unlock the UTXO, it could be a single signature, but it could be several or also some pre-image of a hash or something else. I feel maybe the best for now would be to just have a `get_max_satisfaction_weight` on the UTXO trait to move forward and maybe try to improve later.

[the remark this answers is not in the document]
yancyribbens: I see what you mean now.  Ideally we could just call a method on each UTXO and the precise weight would be returned.  For now I guess it can be approximated with `get_max_satisfaction_weight`.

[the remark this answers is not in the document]
yancyribbens: Just a quick follow up question to check my understanding.  Is the reason we don't know the data required due to the fact that it's due to the fact that it's part of the witness data which isn't stored?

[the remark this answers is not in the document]
yancyribbens: I've been reading a bit more and am still not sure if my understanding is correct.  It seems that a UTXO size isn't known until the party signs and you wouldn't expect the party to provide all possible ways to sign every UTXO.  The follow question assuming this is correct is, why isn't there some type of metadata created when the UTXO is created which can describe the size once it's signed.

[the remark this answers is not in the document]
Xekyo: The output script determines how an input may spend a UTXO. If it's one of the standard scripts, this is sufficient to predict the weight of the input.

- P2PKH: [148 B](https://bitcoin.stackexchange.com/q/48279/5406), 147 B if it has a low-R signature, with a very tiny chance of being even smaller
- P2SH-P2WPKH: [364 WU/91 vB](https://bitcoin.stackexchange.com/q/111392/5406), 363 WU for a low-R signature
- P2WPKH: [272 WU/68 v B](https://bitcoin.stackexchange.com/q/100159/5406), 271 WU for a low-R signature
- P2TR keypath spend: [230 WU/57.5 vB](https://bitcoin.stackexchange.com/q/111395/5406)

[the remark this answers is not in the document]
Tibo-lg: > The output script determines how an input may spend a UTXO. If it's one of the standard scripts, knowledge of the output script is sufficient to predict the weight of the input.
> 
> * P2PKH: [148 B](https://bitcoin.stackexchange.com/q/48279/5406), 147 B if it has a low-R signature
> * P2SH-P2WPKH: [364 WU/91 vB](https://bitcoin.stackexchange.com/q/111392/5406), 363 WU for a low-R signature
> * P2WPKH: [272 WU/68 v B](https://bitcoin.stackexchange.com/q/100159/5406), 271 WU for a low-R signature
> * P2TR keypath spend: [230 WU/57.5 vB](https://bitcoin.stackexchange.com/q/111395/5406)

Yes so way to take advantage of this would be to go with [option 1](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22#discussion_r1068817903) I think.

> Is the reason we don't know the data required due to the fact that it's part of the witness data which isn't stored?

It's because there are many possible scripts that can be used to lock an output which require different inputs of different size.

> The follow question assuming this is correct is, why isn't there some type of metadata created when the UTXO is created which can describe the size once it's signed.

A wallet using a non standard script would probably generate such metadata, but it's not part of the transaction, and I think the closest thing to a standard for this metadata would be an output descriptor.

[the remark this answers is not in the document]
Xekyo: Oh, I did not realize. Do Descriptors specify the input weight of the script they produce?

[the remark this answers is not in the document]
Tibo-lg: So my understanding from [this](https://bitcoin.stackexchange.com/questions/99540/what-are-output-descriptors#answer-99541) and [this](https://docs.rs/miniscript/latest/miniscript/descriptor/enum.Descriptor.html#method.max_satisfaction_weight) is that you can get an upper bound on the witness size that could be required. Though thinking more now for cases like when you spend a Taproot output through the key path you might be overpaying if you had some script path that you don't use,  but not sure if it's worth considering. If it is then we'd need more information probable.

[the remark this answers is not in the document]
yancyribbens: `max_satisfaction_weight` has already been deprecated in favor of `max_weight_to_satisfy` see [here](https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/descriptor/mod.rs#L369).  It sounds like we are OK with just going with descriptor outputs then?  It seems like many wallets are in the process or haven't incorporated descriptors yet.  Also, to answer Xekyo and my previous question, output descriptors store metadata about the derivation path and script key format from my understanding but they don't store metadata about the precise satisfaction size.  Maybe descriptors could be extended to include the satisfaction size of the future so we don't need to use an upper bound estimate?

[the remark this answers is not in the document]
Xekyo: @Tibo-lg: Yeah, the scriptpath would generally be way more expensive. Such behavior may also present a fingerprint that leaks whether a keypath spend has a scriptpath, if people can distinguish what software created the tx.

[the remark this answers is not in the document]
Tibo-lg: > It sounds like we are OK with just going with descriptor outputs then?
I don't know considering that it's still not enough info to compute the input weight properly I'm not sure it's worth the dependency.

>  Such behavior may also present a fingerprint that leaks whether a keypath spend had a scriptpath alternative
Yeah definitely not ideal.

@yancyribbens maybe just go with something abstract or high level for now and try to get the actual coin selection logic implemented, then we can maybe keep brainstorming on the best way to get the input weight info in an issue or something.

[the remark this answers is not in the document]
yancyribbens: @Tibo-lg sure, that makes sense to focus on the selection semantics for now.  Adding weight calculations can come next.

I pushed a new commit that just uses a concrete type for the UTXO and assumes we already know the UTXO weight.  Thoughts?

src/lib.rs
```diff
@@ -17,16 +17,24 @@
...
 
 #[cfg(any(test, feature = "rand"))]
-use rand::{seq::SliceRandom, thread_rng};
+pub use crate::single_random_draw::select_coins_srd;
+
+/// The UTXO values needed to perform coin selection.
+ #[derive(Clone)]
+pub struct Utxo {
+    value: u64,
+    feerate: u64,
```
Tibo-lg, line 29: Fee rate should be a parameter to the selection function

src/lib.rs
```diff
@@ -17,16 +17,24 @@
...
+}
 
-/// Trait that a UTXO struct must implement to be used as part of the coin selection
-/// algorithm.
-pub trait Utxo: Clone {
-    /// Return the value of the UTXO.
-    fn get_value(&self) -> u64;
+impl Utxo {
+    fn get_effective_value(&self) -> u64 {
+        let fee = self.weight * self.feerate;
```
Tibo-lg, line 35: Depending on what is the unit of the fee rate this can be incorrect. Either define the unit or define a `FeeRate` struct that can do conversions.

Tibo-lg (commented):
I'm ok with using a struct though I'm not sure if it gives you more than a trait.

src/lib.rs
yancyribbens: @Tibo-lg I'm not sure I follow you hear.  Can you give me a better example of how the feerate can be incorrect here?

yancyribbens:
@Tibo-lg I think the only downside to using a Trait instead of a struct is the increased compile time.  A monomorphised type has to be constructed at compile time, although the run-time should be the same, so either is probably fine.

src/lib.rs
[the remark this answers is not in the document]
yancyribbens: I moved fee_rate into a new struct called Spend in this [commit](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/63b9fbf84621b892a708261e6e5dc4540e4e842f).  This should help determine which params are needed when a UTXO is spent.  Also, I removed `Clone` in this [commit](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/bd45a30131ad9005978bc4da4bac09c9d8bd6170) since there's an overhead to cloning AFAIK.  Every time clone is called it's making a new copy, so we can instead just pass around pointers.

src/lib.rs
Tibo-lg: The fee rate is often given in sats/vbytes (e.g. [the esplora api](https://github.com/Blockstream/esplora/blob/master/API.md#get-fee-estimates)), but sometimes other units are used, so you need to define the unit you require the use to pass. Having sats/weight unit is quite uncommon.

src/lib.rs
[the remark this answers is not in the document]
Tibo-lg: >I moved fee_rate into a new struct called Spend in this [commit](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/f18b04f9445b2a8305cc2a8f3aa4aa2572cd2d0a). This should help determine which params are needed when a UTXO is spent.

Are there going to be other value in the `Spend` struct? Otherwise it feels like you're just wrapping a single field which doesn't make a lot of sense to me.

> Also, I removed Clone in this [commit](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/917b6b6ced4a4304e0c3b73aa95bda984616d235) since there's an overhead to cloning AFAIK. Every time clone is called it's making a new copy, so we can instead just pass around pointers.

Sure that's fine!

src/lib.rs
yancyribbens: got it, thanks!

src/lib.rs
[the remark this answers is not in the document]
yancyribbens: @Tibo-lg ok, maybe Spend is overkill.  How about just a simple fee module? https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/4c3137b9a638e78aa09dfead4b8452f351beee29#diff-f4eff9ed318d5e5d7acac95241be35055cf36968693da8a60c16b8b53e97fdd4

I assume we'll want to use the fee helper functions for other algorithms which is why I put it in a separate module.

src/fee.rs
```diff
@@ -0,0 +1,50 @@
...
+
+//One vbyte is equal to four weight units.
+#[cfg(any(test, feature = "rand"))]
+fn weight_to_vbytes(weight: u64) -> u64 {
+    weight / 4
+}
+
+#[cfg(any(test, feature = "rand"))]
+pub fn get_fee(fee_rate: u64, utxo: &impl Utxo) -> u64 {
+    weight_to_vbytes(utxo.get_weight()) * fee_rate
```
Tibo-lg, line 14: If you want the library to be robust you might want to do a checked_mul and error if it overflows.

src/lib.rs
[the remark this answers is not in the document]
Tibo-lg: Sure that's fine!

src/fee.rs
yancyribbens: I think this is debatable.  The problem with `checked_mul` is that it will return `None` to the calling application instead of a panic which will return a stack trace with potentially useful information to debug.  Another way to look at it is that wouldn't a multiplication that's greater than u64 max be a fatal error?  If the error should not be fatal then I could see wanting to use `checked_mul` and return an `Error` type with a useful message I suppose.

yancyribbens: I think I'd prefer using [debug_assert](https://doc.rust-lang.org/std/macro.debug_assert.html) to panic possibly with a message.  You're right though that wrapping silently if the result exceeds u64 max is a probably unexpected, and in release mode this could happen, unlike debug mode.

Tibo-lg: My thinking is that as much as possible I expect the library that I use not to panic, but to error if I give them invalid arguments (which would be the case here IMHO with a fee_rate that would be too high). I feel whether the error is fatal or not is for the caller to decide.

yancyribbens: Ok I see your point.  I suppose it's possible the application allows user input, and doesn't validate the input and then passes extreme values to this lib.  In such a case an error could be returned to prompt the user for a lower fee_rate value (assuming the calling application handles the error).  The other option is just to document that in the library that the application will panic if the values exceed u64 max so the application should validate and send sane values.  Anyway, if you feel that strongly about returning an error that makes some sense and I can add it.

Tibo-lg: I don't feel strongly about it, documenting it would also work but it might be a bit awkward because you can't really give a value since it also depends on the weight.

yancyribbens: Since `fee_rate` could be user input, I've changed my mind and agree that the application shouldn't crash if a user inputs a crazy value.  It's better if the application validates the input but the lib can also help guard against such occurrences.

yancyribbens: @Tibo-lg resolved by this commit https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/7c6b6343330a600a830c5cf800b4f2a9a875194f

yancyribbens: the error propagation of https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/c9cf487127c2ed8b7f10cc0415cffc15eb25c103 is a bit weird, although I'm not sure of a better way and still using an iterator.  Using an iterator is preferred for performance probably but maybe a standard `for each` would be easier to grok.

Tibo-lg: Yeah I would personally go for code readability first and then consider optimizing later once everything is implemented so you can benchmark to see how big of a difference it makes. Also I wouldn't recommend adding the `num` and `error` dependencies, in general the crates within `rust-bitcoin` try to use as few dependencies as possible.

yancyribbens: @Tibo-lg ok I've removed the `num` and `error` deps, also I got ride of the iterator `take_while` for now: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/bf7316bec83e0ade6a2d95a7f9992e26ace8e383

src/single_random_draw.rs
yancyribbens: Added RNG here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/9ea6fab1c65259fc98667d7c0400a71142158275

yancyribbens: @Xekyo sorry it's taken me a while to get back to you on this.  I believe this commit should address your comment: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/542e2b236026f0b5591c5f0f99b59def7f34795c

yancyribbens:
I believe this PR is ready to merge.  Any objections @Tibo-lg ?

src/fee.rs
```diff
@@ -0,0 +1,25 @@
...
+    vbytes * fee_rate
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::fee::*;
+
+    #[test]
+    fn get_effective_value() {
+        let fee_rate = 10;
```
Xekyo, line 34: If you are going with integers for the feerate, I would recommend using `ṩ/kvB` as the base unit, since it allows you to be more precise than just full satoshi increments. That way you can create transactions with feerates at `1.103 ṩ/vB`, instead of needing to pay `2 ṩ/vB` if `1 ṩ/vB` is insufficient for the priority you want to imbue the transaction with.

Xekyo: I was also about to ask about how you avoid dust outputs, but then I saw the last commit. :+1:

src/fee.rs
```diff
@@ -0,0 +1,66 @@
...
+#[cfg(any(test, feature = "rand"))]
+#[allow(arithmetic_overflow)]
+pub fn get_fee_vbytes(fee_rate: u64, weight: u64) -> Result<u64, LibError> {
+    if fee_rate == 0 {
+        Err(LibError::FeeRateSize)
+    } else if weight == 0 {
+        Err(LibError::WeightSize)
+    } else {
+        let vbytes = weight / 4;
+        let fee = fee_rate.wrapping_mul(vbytes);
```
Tibo-lg, line 16: why don't you just use `checked_mul`?

src/single_random_draw.rs
```diff
@@ -0,0 +1,199 @@
+//! This library provides efficient algorithms to compose a set of unspent transaction outputs
+//! (UTXOs).
+
+#[cfg(any(test, feature = "rand"))]
```
Tibo-lg, line 4: Maybe put the feature gate on the module itself?

src/fee.rs
yancyribbens: `checked_mul` is part of the `num` crate: https://docs.rs/num/latest/num/trait.CheckedMul.html.  As you said before, it's best to minimize external dependencies.  Is there a different `checked_mul` that I don't know about?

Tibo-lg: https://doc.rust-lang.org/std/primitive.u64.html#method.checked_mul

yancyribbens: thanks, done.

src/single_random_draw.rs
yancyribbens: good idea.  done.

src/fee.rs
yancyribbens: > If you are going with integers for the feerate, you could consider using ṩ/kvB as the base unit

Thanks, that's a good point.  Wouldn't it be better to use a float for rate?

Otherwise, [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/files#diff-f4eff9ed318d5e5d7acac95241be35055cf36968693da8a60c16b8b53e97fdd4R14) would be: 

`let kv_bytes = weight / 4 /1024;`
which means we lose a lot of precision with int division OR kvbytes is a float.  It doesn't work to have kvbytes be a float and feerate as an int then.

Xekyo: We don’t use “kibivirtualbytes”, but “kilovirtualbytes“:

    1 MvB = 1000 kvB = 1 000 000 vB
    
We never use floats for consensus relevant calculations in Bitcoin Core to avoid any sort of rounding issues. YMMV.

yancyribbens: How about something like this: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22/commits/d1d38b16611e45a57ded93daa715126216156b84

```
/// Takes as input the fee_rate in $ per Kilovirtualbytes
/// and a size in Kilovirtualbytes and returns the fee amount.
///
/// * fee_rate: $ per kilovirtualbytes $KvB
/// * size: kilovirtualbytes KvB
/// * fee: $
```

yancyribbens: I just noticed this PR https://github.com/rust-bitcoin/rust-bitcoin/pull/1627.  Maybe it would be better to build off `FeeRate` as defined by Rust Bitcoin.

Tibo-lg: SGTM

Xekyo: > How about something like this: [d1d38b1](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/commit/d1d38b16611e45a57ded93daa715126216156b84)
> 
> ```
> /// Takes as input the fee_rate in $ per Kilovirtualbytes
> /// and a size in Kilovirtualbytes and returns the fee amount.
> ///
> /// * fee_rate: $ per kilovirtualbytes $KvB
> /// * size: kilovirtualbytes KvB
> /// * fee: $
> ```

It’s not a dollar symbol but an “s” with a dot above and below: ṩ (u+1e69). The symbol for kilo is “k”, not “K”, the latter is the symbol for Kelvin, so it’s kvB, and that makes the feerate ṩ/kvB (which without the slash would be a multiplication rather than a division).

yancyribbens:
> Hey, this keeps popping up in my notifications but doesn’t seem to be ready for review. I’ll stop following notifications here for the time being. Please feel free to ping me here again when this PR is ready for review.

Sorry for all the noise.  I think it's about ready now, however there seems to be some unrelated failure with the nightly toolchain I'm trying to figure out.  Anyway, I think at this point I'll just close this PR and open a fresh one once I figure out what's going on with CI.

yancyribbens:
@Xekyo a new PR is now open here https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/23 - cheers.

PR #23, opened 2023-07-03
Feature/add srd selection module

yancyribbens:
This is a new PR for adding Single Random Draw module.  The first PR was here https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/22 however numerous changes have been made so much of the conversation there has become stale.  I've done a lot of work also on rust-bitcoin upstream so that this feature is as streamlined.  There is one additional upstream change to rust-bitcoin that I haven't not created a PR for yet [here](https://github.com/rust-bitcoin/rust-bitcoin/compare/master...yancyribbens:rust-bitcoin:add-option-param-to-predict_weight) as I wanted to make sure that it is needed.


yancyribbens:
cc @Tibo-lg @Xekyo

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+    fee_rate: FeeRate,
+    rng: &mut R)
+-> Result<Option<Vec<TxOut>>, LibError> {
+    outputs.shuffle(rng);
+
+    let mut selected_effective_value: Amount = Amount::ZERO;
+    let mut i = 0;
+    let mut result = Vec::new();
+
+    // The required size of a change output must be at least
```
Tibo-lg, line 30: What if a change output is not required?

src/lib.rs
```diff
@@ -9,25 +9,36 @@
...
-/// Trait that a UTXO struct must implement to be used as part of the coin selection
-/// algorithm.
-pub trait Utxo: Clone {
-    /// Return the value of the UTXO.
-    fn get_value(&self) -> u64;
-}
+#[cfg(any(test, feature = "rand"))]
+use rand::thread_rng;
+
+#[cfg(any(test, feature = "rand"))]
```
Tibo-lg, line 28: instead of having to put this everywhere, you can force the `rand` feature to be used in tests by putting this under the dev-dependencies in the cargo file:
```
rust-bitcoin-coin-selection = {path = ".", features = ["rand"]}
```

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+    // larger than the target by CHANGE_LOWER.  Therefore,
+    // CHANGE_LOWER + target becomes the new lower bound for the
+    // selection.
+    while selected_effective_value < target + CHANGE_LOWER && i < outputs.len() {
+        let output: &TxOut = &outputs[i];
+        let value: Amount = output.value;
+
+        let outputs = outputs.iter().map(|txout| txout.script_pubkey.len());
```
Tibo-lg, line 38: is it normal that you're taking the length for all outputs instead of just the ones being selected?

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let weight: Weight = transaction::predict_weight(
+            None::<Vec<transaction::InputWeightPrediction>>,
```
Tibo-lg, line 41: nit: use `_` instead of `transaction::InputWeightPrediction`

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+            Some(outputs),
+        );
+
+        let fee: Option<Amount> = fee_rate.checked_mul_by_weight(weight);
+
+        match fee {
+            Some(f) => {
+                let effective_value = value - f;
```
Tibo-lg, line 49: the effective value seem to be computed using the fee for all outputs, which seems incorrect to me. I guess it should only use the fee for the added input?

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+    // selection.
+    while selected_effective_value < target + CHANGE_LOWER && i < outputs.len() {
+        let output: &TxOut = &outputs[i];
+        let value: Amount = output.value;
+
+        let outputs = outputs.iter().map(|txout| txout.script_pubkey.len());
+
+        let weight: Weight = transaction::predict_weight(
+            None::<Vec<transaction::InputWeightPrediction>>,
+            Some(outputs),
```
Tibo-lg, line 42: Shouldn't it be checking the fees for inputs rather than outputs? (selecting an additional utxo will add an input to the transaction, but maybe I'm misunderstanding how this function is supposed to work).

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    let mut selected_effective_value: Amount = Amount::ZERO;
+    let mut i = 0;
+    let mut result = Vec::new();
+
+    // The required size of a change output must be at least
+    // larger than the target by CHANGE_LOWER.  Therefore,
+    // CHANGE_LOWER + target becomes the new lower bound for the
+    // selection.
+    while selected_effective_value < target + CHANGE_LOWER && i < outputs.len() {
```
Tibo-lg, line 34: nit: I'd find it more readable to use a for loop over the outputs and break within if target is reached, but might be a question of taste.

Tibo-lg (commented):
Did a first pass, I might be misunderstanding but some things look strange to me.

src/single_random_draw.rs
yancyribbens: good catch.  This should be the set of outputs that have been selected AKA the result vec.  Every time a new utxo is added `result` predict_weight can recalculate the weight total then.

src/errors.rs
```diff
@@ -0,0 +1,24 @@
+/// Error types.
+
+use bitcoin::FeeRate;
+use bitcoin::Weight;
+use std::{error::Error, fmt};
+
+#[derive(Debug)]
+pub enum LibError {
```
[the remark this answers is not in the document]
yancyribbens: That's fine also I think.  I can rename this to `Error`.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
 # 0.1.0 - 2022-12-17
 
 * Initial release.
+
+# 0.2.0 - 2023-06-03
+
+- Add Single Random Draw module and a basic error type.
+- Numerous errors where found in the BnB selection Algorithm.  Removed BnB selection Algorithm, tests and benchmarks until a new version is developed.
```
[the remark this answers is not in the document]
yancyribbens: My intuition was to remove it until the issues have been fixed.  Although, if it's better to leave it until the next release when an improved version is available, that also works.

src/single_random_draw.rs
yancyribbens: I modeled this after the core SRD selection, and I believe the idea is that SRD is a fallback when no solution can be found that does not require a change output.  Perhaps it's useful to update the comment to explain that?

src/single_random_draw.rs
yancyribbens: I think since reading Edgar Dijkstra paper Gotos considered harmful, I try to avoid breaks or returns if possible.  Also Rust borrows from Ruby a lot with the implicit return at the end of functions, and it's popular to prefer that instead of writing return or break to a return.  Although, if a `for`  loop is more commonly used, and others find that easier to read, I'm fine with that instead.  In the end, I agree it's mostly a matter of taste as you say :)

src/single_random_draw.rs
yancyribbens: hmm that's a good point, and I think you are correct.  The owner of the outputs will transition them to be inputs for the transaction, and so, when included in a transaction, they are actually now inputs instead of outputs, which changes the weight calculation.

src/single_random_draw.rs
yancyribbens: > Shouldn't it be checking the fees for inputs rather than outputs?

Well, based on your previous comment it should be the fee for all inputs?  Also, I think this will be corrected if the [line here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/23/files/3ab168d23d2ab2c52bdda669ae6eb5fcd99df1de#diff-ce2c6b6b3132bf7b26a80fc104adfdeb0eb023b9bfd487c0da95d83830a73a32R38) is changed to be the selected "inputs".

src/errors.rs
```diff
@@ -0,0 +1,24 @@
...
+    Multiplication(Weight, FeeRate),
+}
+
+#[cfg(any(test, feature = "rand"))]
+impl Error for LibError {}
+
+#[cfg(any(test, feature = "rand"))]
```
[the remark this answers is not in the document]
yancyribbens: I think I did that to avoid compiler warning that it's not otherwise used if the `rand` flag is not used.  If there are more selection algorithms such as BnB than this feature flag can be removed and there will no longer be compiler warnings.

src/single_random_draw.rs
yancyribbens: ack

src/single_random_draw.rs
```diff
@@ -0,0 +1,166 @@
...
+        let output: &TxOut = &outputs[i];
+        let value: Amount = output.value;
+
+        let outputs = outputs.iter().map(|txout| txout.script_pubkey.len());
+
+        let weight: Weight = transaction::predict_weight(
+            None::<Vec<transaction::InputWeightPrediction>>,
+            Some(outputs),
+        );
```
Xekyo, line 43: Shouldn’t this calculate the input weight on basis of the specific `output`, because you could have outputs of different types?

src/single_random_draw.rs
Xekyo: Agree with @Tibo-lg, the effective value of an input is the UTXO’s amount minus the cost of spending that UTXO at a given feerate. Also be sure to skip UTXOs with a negative effective value in selection.

src/single_random_draw.rs
yancyribbens: I updated this to calculate the specific weight of every output so far selected.  Is this what you had in mind?

src/single_random_draw.rs
yancyribbens: this has been fixed now I think.

src/single_random_draw.rs
yancyribbens: I've update this so the effective_value is now based on the sum of the utxos so far selected.

src/single_random_draw.rs
Xekyo: I meant that “effective_value” is a per-input property. It is calculated from the UTXO’s amount, the corresponding input’s weight, and the current feerate. It must be evaluated on a per-input basis because the effective value of individual UTXOs can turn negative if they are worth a small amount and the feerate is sufficiently high. At that point, you’d want to exclude them, because they don’t help you progress towards funding the transaction, but rather increase the total amount that needs to be provided by other inputs.

yancyribbens: > ..the corresponding input’s weight..

I'm not certain what is meant by the "corresponding input" here.  Is this the input that's to be constructed which consumes the UTXO?

murchandamus: Yes

src/errors.rs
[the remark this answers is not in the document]
yancyribbens: Changed to `Error` now

src/single_random_draw.rs
yancyribbens: > Also be sure to skip UTXOs with a negative effective value in selection.

Amount is of type U64 which can't be negative, so rust-bitcoin produces a panic if this happens.

src/single_random_draw.rs
yancyribbens: Why is this not expressed in the bitcoin-core version?  AFAICT core only computes the weight of a group of UTXOs: https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L220

src/single_random_draw.rs
yancyribbens: changed this to a for loop similar to the core implementation.

src/single_random_draw.rs
yancyribbens: Perhaps I'm still misunderstanding.  I've updated the computation to be on a per-utxo basis similar to how core works.

src/single_random_draw.rs
Tibo-lg: Then wouldn't it be better to do a `checked_sub`? It feels the `> 0` check is meaningless for now.

src/errors.rs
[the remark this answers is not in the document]
Tibo-lg: I don't get any more warning if I remove these `cfg` flags. Though there is an unused warning on the unused `FeeRate` in the `lib.rs` file if compiling without `rand` feature.

CHANGELOG.md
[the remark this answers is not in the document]
Tibo-lg: I don't have a strong opinion, it depends if you want to publish a `0.2.0` version with just srd or not. If not I'd keep that out of the changelog, if yes then you can keep it.

src/single_random_draw.rs
murchandamus: Bitcoin Core groups all UTXOs with the same scriptPubKey to form an OutputGroupt to opportunistically mitigate address reuse: if multiple payments have been received to the same address, they will generally be spent together instead of in multiple separate transactions that would then be revealed to have been sent by the same party.

yancyribbens: Cool, I'll take a look more into UTXO groups in core.  Maybe that's something that could be added to Rust Bitcoin?

clippy.toml
```diff
@@ -1 +1 @@
-msrv = "1.41.1"
+msrv = "1.48.1"
```
yancyribbens, line 1: this should be 1.48.0

src/errors.rs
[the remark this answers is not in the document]
yancyribbens: oops seems you are right.  fixed.

src/single_random_draw.rs
yancyribbens: Good idea, thanks! done.

CHANGELOG.md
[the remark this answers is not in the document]
yancyribbens: maybe best to only add SRD now and remove BnB iin a future PR and simplify the changelog.

src/lib.rs
yancyribbens: Thanks!  Good idea.

yancyribbens: removed all feature gates and made rand default for dev mode

src/single_random_draw.rs
yancyribbens: This will now skip any utxos with a negative effective value.

src/single_random_draw.rs
yancyribbens: Fair enough.  I'm not understanding what is missing currently.  predict_weight() can take a set of inputs and outputs,  Currently I'm passing each utxo to predict_weight() and calculating the total.  Is this not correct?  I think predict_weight() should be used however to calculate the entire set of candidate utxos instead of individually adding to the total.

yancyribbens: s/calculating the total/adding to the total/g

murchandamus: When you build a transaction, you must select sufficient funds for a) creating the recipient outputs, b)  fees for the transaction to have the intended queuing priority, and c) a sufficient budget for a change output.

You know in advance what amount the recipient outputs sum up to, and how much budget it takes to create a non-dust change output of a known type at a specific feerate—you can include these amounts in the selection target.

However, the fees for the inputs depend on the number and type of inputs that get selected. You _cannot_ include these in the selection target, because then your selection target _shifts_ every time you select or deselect an input, depending on the weight of that input and the current feerate. It’s really hard to hit a moving target.

This problem is mitigated by keeping track of the _effective values_ of the inputs you have selected rather than their full amount—by deducting the cost of spending the UTXO from the UTXO’s amount, you can account for the net amount that selecting the input moves you closer to the selection target, because _the input has “already paid for itself”_ when you select it. That’s the point of UTXO’s “effective value” and why it’s calculated from the specific UTXO’s corresponding input weight and the current feerate.

murchandamus: I’m a bit at a loss here. It seems that I’m not finding the right words to explain the basic constraints and goals of Coin Selection to you. You may find it useful to read this [Stack Exchange answer](https://bitcoin.stackexchange.com/a/114847/5406) as well as the 3rd and 5th chapter of my [Master thesis](https://murch.one/erhardt2016coinselection.pdf).

yancyribbens: Thanks, I think I understand now.  From here the difficult part is to figure out how to construct the corresponding input for each utxo.  I suppose the simplest thing is just require three additional arguments to the selection algorithm ( [script_sig,](https://github.com/rust-bitcoin/rust-bitcoin/blob/1318ff88e525cd5ef3afd169c8dfd4934271e58c/bitcoin/src/blockdata/transaction.rs#L188) [sequence](https://github.com/rust-bitcoin/rust-bitcoin/blob/1318ff88e525cd5ef3afd169c8dfd4934271e58c/bitcoin/src/blockdata/transaction.rs#L193) and [witness](witness)) for which every corresponding input will be built with.  If my understanding is correct, there will be `n` inputs and two outputs (destination output, change output) and then the remainder of the fields to calculate the weight from.

yancyribbens: I've yet to re-write the tests.  However this is what I was thinking should work.  Basically construct the transaction, get the weight and ensure that the fee can be paid for by the change_output, if not keep adding inputs to the transaction.

```
    utxos.shuffle(rng);

    let zipped = utxos.iter().zip(inputs.iter());
    let mut inputs:Vec<TxIn> = Vec::new();
    let mut value = Amount::ZERO;

    for (u, i) in zipped {
        value += u.value;
        inputs.push(i.clone());

        if value > target + CHANGE_LOWER {

            // TODO use chedcked_sub
            let change_amount = value - target - CHANGE_LOWER;

            let change_output = TxOut {
                value: change_amount,
                script_pubkey: script_pubkey.clone()
            };

            let destination_output = TxOut {
                value: target,
                script_pubkey: script_pubkey.clone()
            };
            
            transaction.input = inputs.clone();
            transaction.output = vec![change_output, destination_output];
                
            let weight = transaction.weight();
            let transaction_fee: Option<Amount> = fee_rate.checked_mul_by_weight(weight);
    
            match transaction_fee {
                Some(f) => {
                        
                    // there is not enough to pay the fee out of the change output
                    // without creating a dust output.  Remove the last input so
                    // another permutation can be tried.
                    if change_amount < f - CHANGE_LOWER{
                        inputs.pop();
                        transaction.input = inputs.clone();
                        value -= u.value;
                    } else {
                        return Ok(Some(transaction.clone()))
                    }

                },
                None => return Err(Error::Multiplication(weight, fee_rate)),
            }
        }
    }

    Ok(None)
}
```

murchandamus: As far as I can tell, this will skip UTXOs with negative effective value, but still accounts for inputs with the whole value rather than the effective value towards the target. If my understanding of your code is right, 

        value += u.value;
        
should read
        
        value += u.value - input_fee;

I suggest that you implement some tests that check whether your code selects sufficient funds to both pay the fee and still exceed `target + CHANGE_LOWER` with its selection.

yancyribbens: Thanks!  I've updated the tests and the logic as suggested.  Also I've also added an assertion to test that the change amount is greater than the dust limit once fees are subtracted [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/23/files#diff-ce2c6b6b3132bf7b26a80fc104adfdeb0eb023b9bfd487c0da95d83830a73a32R146)

src/single_random_draw.rs
yancyribbens: Thanks, I believe this has been resolved now.

src/single_random_draw.rs
yancyribbens: I believe this has been resolved now.

src/single_random_draw.rs
```diff
@@ -0,0 +1,432 @@
...
+
+            let weight: Weight = transaction.weight();
+            let transaction_fee: Option<Amount> = fee_rate.checked_mul_by_weight(weight);
+
+            match transaction_fee {
+                Some(f) => {
+                    // if the fee is very high then change_amount - fee could be negative.
+                    // This will therefore be None in such a case so that more inputs are
+                    // selected.
+                    let effective_value = change_amount.checked_sub(f);
```
murchandamus, line 101: Another way to handle this is by including the fee for the change output in the target for the change budget, see e.g. https://github.com/bitcoin/bitcoin/pull/27846

CHANGELOG.md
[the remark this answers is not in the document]
yancyribbens: reverted this change for now.

src/single_random_draw.rs
murchandamus: I meant that if the “change budget” is supposed to cover fees, but it is possible for the change budget to be smaller than the fee here and you need to go back to select more inputs, then you should probably set the change budget such that it already explicitly includes the fees.

yancyribbens: About the paramater [change_fee](https://github.com/murchandamus/bitcoin/blob/1771daa815ec014276cfcb30c934b0eaff4d72bf/src/wallet/test/fuzz/coinselection.cpp#L68) that gets passed in to SRD; What I think you are suggesting is just add the amount for creating a change_output and add that to the budget.  so  `if value >= target + CHANGE_LOWER {` becomes `if value >= target + CHANGE_LOWER + change_fee`.  However that doesn't preclude the possibility for `let effective_value = change_amount.checked_sub(f);` to be negative if fees are extreme?

yancyribbens: I created a draft commit https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/23/commits/d6d4d4d7373cb02424f8a1e0d9f8773c38f86f79 to try out your suggestion.  However there is a failing test now since adding change_fee to the budget doesn't preclude the fee from being larger than the change output.

yancyribbens: I'm going to undo the draft commit since it's obviously broken.  I'll take a look more into cores implementation since there is a detail I think I'm missing.  Another item on the short list will be to add fuzz testing.  Unless I hear otherwise I'll mark this conversation resolved.

yancyribbens: > Another way to handle this is by including the fee for the change output in the target for the change budget

I made some minor change to calculate the fee for the outputs and transaction as part of the change budget.

yancyribbens:
Is there any other change requests or improvement ideas?

yancyribbens:
I added a commit to check for max transaction weight similar to core: https://github.com/bitcoin/bitcoin/blob/f4f1d6d230dd390f0524b8852b8cfc912d006958/src/wallet/coinselection.cpp#L225

src/single_random_draw.rs
```diff
@@ -0,0 +1,450 @@
...
+    let input_weight = transaction::predict_weight(Some(vec![weight_predict]), None::<Vec<_>>);
+    let input_fee: Option<Amount> = fee_rate.checked_mul_by_weight(input_weight);
+
+    match input_fee {
+        Some(f) => Ok(utxo.value.checked_sub(f)),
+        None => Err(Error::Multiplication(input_weight, fee_rate)),
+    }
+}
+
+/// Calculate the fee for a transaction with two UTXOs and no inputs.
```
murchandamus, line 46: Should “a transaction with two UTXOs and no inputs” be “a transaction with two outputs and no inputs”?

src/single_random_draw.rs
```diff
@@ -0,0 +1,450 @@
...
+fn calculate_fee(
+    script_pubkey: &ScriptBuf,
+    fee_rate: FeeRate,
+    transaction: &mut Transaction,
+) -> Result<Amount, Error> {
+    let output: TxOut = TxOut { value: Amount::ZERO, script_pubkey: script_pubkey.clone() };
+
+    let output_two = output.clone();
+
+    transaction.output = vec![output, output_two];
```
murchandamus, line 63: 1. It’s not clear to me why this assumes that both outputs have the same weight. What happens for example if your recipient output is an OP_RETURN output or a wrapped segwit output and your change is P2WPKH?
2. It’s not clear to me why this assumes always two outputs. What happens if you are trying to send all money in your wallet or to send to multiple recipients?

src/single_random_draw.rs
```diff
@@ -0,0 +1,488 @@
...
+/// /// * `inputs` - the order corresponds to the ordering of the UTXOs where the first input consumes the first UTXO and so on.
+/// /// * `transaction` - the transaction to populate with resulting inputs and outputs.
+/// /// * `script_pubkey` - needed to compose the transaction outputs.
+/// /// * `rng` - used primarily by tests to make the selection deterministic.
+///
+pub fn select_coins_srd<R: rand::Rng + ?Sized>(
+    target: Amount,
+    fee_rate: FeeRate,
+    utxos: &mut [TxOut],
+    inputs: &mut [TxIn],
```
Tibo-lg, line 103: You seem to be assuming that the `TxIn` already has the witness elements filled in, but that's usually not the case. I think we already had this discussion a while ago, but you probably need to either have the user provide the maximum witness length or use miniscript so that you can compute it.

src/single_random_draw.rs
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        transaction.input.push(input.clone());
+
+        if transaction.weight() >= MAX_TRANSACTION_SIZE {
+            return Err(Error::MaxTransactionWeight(transaction.weight()));
+        }
+
+        if value >= threshold {
+            let change_amount = target - fee;
```
Tibo-lg, line 141: Probably a rare case but can't this panic if fee > target?

src/single_random_draw.rs
```diff
@@ -0,0 +1,488 @@
...
+/// ///
+/// /// * `target` - target value to send to recipient.
+/// /// * `fee_rate` - ratio of transaction amount per size.
+/// /// * `utxos` - UTXOs from which to sum the target amount.
+/// /// * `inputs` - the order corresponds to the ordering of the UTXOs where the first input consumes the first UTXO and so on.
+/// /// * `transaction` - the transaction to populate with resulting inputs and outputs.
+/// /// * `script_pubkey` - needed to compose the transaction outputs.
+/// /// * `rng` - used primarily by tests to make the selection deterministic.
+///
+pub fn select_coins_srd<R: rand::Rng + ?Sized>(
```
Tibo-lg, line 99: The API looks more like `fund_transaction` than `select_coins`.

src/single_random_draw.rs
yancyribbens: yes, good point.  thanks!

src/single_random_draw.rs
yancyribbens: I see your point.  Although I needed a way to compute the weight of the transaction that the coins might need to be included in (transaction headers).  Instead I could just expect that the weight is passed as an argument to select_coins.

src/single_random_draw.rs
yancyribbens: Is it unreasonable to assume that for each `UTXO` passed, that the wallet software using the interface would create a corresponding `TxIn`?  Maybe there is too much of a performance penalty?

> or use miniscript so that you can compute it.

Are you suggesting that miniscript can compute the size of the witness elements given a UTXO?  I need to look more into how miniscript works.  My understanding is just that it's a simplified version of bitcoin script.

src/single_random_draw.rs
```diff
@@ -0,0 +1,488 @@
...
+        transaction.input.push(input.clone());
+
+        if transaction.weight() >= MAX_TRANSACTION_SIZE {
+            return Err(Error::MaxTransactionWeight(transaction.weight()));
+        }
+
+        if value >= threshold {
+            let change_amount = target - fee;
+
+            let change_output: TxOut =
```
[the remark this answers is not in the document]
yancyribbens: That should be possible since:

`let threshold = target + CHANGE_LOWER + fee;`

Threshold is greater than the target by change_lower, so any selection exceeds the target by at least the dust limit.

maybe it's worth putting an assert in to make sure?

src/single_random_draw.rs
Tibo-lg: > Is it unreasonable to assume that for each UTXO passed, that the wallet software using the interface would create a corresponding TxIn? Maybe there is too much of a performance penalty?

It's not only a question of performance, but it's just not very practical. I personally wouldn't use a library that forces me to generate the witnesses for all of my UTXOs before they are even selected. I'd be fine with manually passing the maximum witness weight though.

> Are you suggesting that miniscript can compute the size of the witness elements given a UTXO? I need to look more into how miniscript works. My understanding is just that it's a simplified version of bitcoin script.

Yes miniscript can tell you what is the maximum witness length, see https://docs.rs/miniscript/latest/miniscript/descriptor/enum.Descriptor.html#method.max_weight_to_satisfy

src/single_random_draw.rs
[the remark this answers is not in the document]
Tibo-lg: Ah I had forgotten that you wanted this to always use a change output because it's a fallback. I still don't really see why not handle the case where you just end up lucky and don't need a change output but not a big deal.

src/single_random_draw.rs
Tibo-lg: Personally I'd find more useable an interface that just takes a target value which excludes the fee for selected inputs but includes everything else. For example in rust-dlc the fee for the fund transaction is split between both parties, so this interface would be difficult to use.

src/single_random_draw.rs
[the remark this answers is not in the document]
yancyribbens: It's probably worth adding the case where no change output is required, that way it's more modular and doesn't depend on BnB or another algo to be run first.

src/single_random_draw.rs
yancyribbens: I see.  Sure we can try that and see how it looks.

yancyribbens: > Personally I'd find more usable an interface that just takes a target value which excludes the fee for selected inputs but includes everything else.

Besides calculating the fee, I also use the transaction weight to check that the max transaction weight hasn't been exceeded [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/23/files#diff-ce2c6b6b3132bf7b26a80fc104adfdeb0eb023b9bfd487c0da95d83830a73a32R133)

Would passing in a "weight so far" as an argument be possible?

src/single_random_draw.rs
yancyribbens: fixed

src/single_random_draw.rs
yancyribbens: resolved

src/single_random_draw.rs
yancyribbens: Besides, Witness, it looks like [ScriptBuf](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/transaction.rs#L187 ) is also an unknown size since it's a [vec](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/script/owned.rs#L35).  Maybe it's better to pass in the maximum input size?

src/single_random_draw.rs
Tibo-lg: I'm not sure it's a useful enough feature to warrant a more complicated API personally

yancyribbens: I think there is a miss communication here.  You are suggesting incorporating the fee for other parts of the transaction into the target amount.  How then do we know what the weight so far is so that we can guard against the inputs exceeding the weight max when they are selected?

Tibo-lg: I think users can check by themselves after, but if you feel like it's better to have it then go with it!

yancyribbens: Good point.  Instead of having each algo check that max weight hasn't been exceeded, it could just be checked afterwords.

yancyribbens: Ok I gutted all the transaction stuff.

yancyribbens: @murchandamus I'm not sure if you have an opinion but @Tibo-lg thought to just include the fee in the target instead of how core does it here: https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L204

src/single_random_draw.rs
yancyribbens: this is outdated now.  Marking as resoled.

src/single_random_draw.rs
yancyribbens: It seems like there are no further comments or objections here.  Marking as resolved.

src/single_random_draw.rs
yancyribbens: I'll need to do some more research here since there isn't much else for me to go on here.

src/single_random_draw.rs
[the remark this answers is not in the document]
yancyribbens: this comment is outdated since I've removed the transaction building.  Therefore change_outputs are not created here (or lack of change outputs).  marking resolved.

src/single_random_draw.rs
yancyribbens: Ok, so it looks like for each TxOut, we compute the corresponding TxIn as:

```
input_weight = (32 + 4 + 4) * 4 + input.max_weight_to_satisfy();
```

At first glance however at miniscript, it seems that miniscript works on a transaction, not an individual input.   Would it work to pass in `max_satisfaction_weight` for the TxIn that corresponds to each UTXO?  That appears to be how the BDK implementation works: https://github.com/bitcoindevkit/bdk/blob/feafaaca31a0a40afc03ce98591d151c48c74fa2/crates/bdk/src/wallet/coin_selection.rs#L716

Tibo-lg: > At first glance however at miniscript, it seems that miniscript works on a transaction, not an individual input.

No you can generate descriptors that represent spending conditions for a single output.

> Would it work to pass in max_satisfaction_weight for the TxIn that corresponds to each UTXO?

Yes, that's what I was referring to when I wrote:
> I'd be fine with manually passing the maximum witness weight though.

yancyribbens: I see.  There really isn't very much when I search for "satisfaction weight".  So, the satisfaction weight is the base input weight plus the size of the spending conditions of the corresponding UTXO.  I'll take a closer look at miniscript because it would be nice if the selection algo could compute the "satisfaction weight" instead of needing to be passed as a param.

Tibo-lg: AFAIK satisfaction weight is the maximum weight of the parameters that need to be given to satisfy the spending condition, see e.g. how they compute the `P2WPKH_SATISFACTION_SIZE_CONSTANT`: https://github.com/bitcoindevkit/bdk/blob/master/crates/bdk/src/wallet/coin_selection.rs#L704

And tbh if I were you I'd go first with something like that, get a something usable and then maybe later miniscript if it feels like it would add value.

yancyribbens: > AFAIK satisfaction weight is the maximum weight of the parameters that need to be given to satisfy the spending condition, see e.g. how they compute the

Yeah, after further inspection of the bitcoin source [here](https://github.com/bitcoin/bitcoin/blob/master/src/script/miniscript.h#L38) it looks like satisfaction is used to describe an argument passed to the stack that "satisfies" the requirement.  So the satisfaction weight is the size of all the required params that get pushed on the stack.

src/single_random_draw.rs
```diff
@@ -0,0 +1,251 @@
...
+    let weight = match checked_weight {
+        Some(w) => w,
+        None => return Err(Error::AdditionOverflow(satisfaction_weight, TXIN_BASE_WEIGHT)),
+    };
+
+    let input_fee: Option<Amount> = fee_rate.checked_mul_by_weight(weight);
+
+    match input_fee {
+        Some(f) => Ok(weighted_utxo.utxo.value.checked_sub(f)),
+        None => Err(Error::MultiplicationOverflow(satisfaction_weight, fee_rate)),
```
Tibo-lg, line 37:
```suggestion
    let weight = satisfaction_weight
        .checked_add(TXIN_BASE_WEIGHT)
        .ok_or_else(|| Error::AdditionOverflow(satisfaction_weight, TXIN_BASE_WEIGHT))?;

    let input_fee = fee_rate
        .checked_mul_by_weight(weight)
        .ok_or_else(|| Error::MultiplicationOverflow(satisfaction_weight, fee_rate))?;

    Ok(weighted_utxo.utxo.value.checked_sub(input_fee))

```

Tibo-lg (commented):
Small neat on code that can be simpler IMHO but otherwise looks like it's actually usable.

yancyribbens:
@Tibo-lg thanks for all the suggestions and the review!  I used your code snippet  with `ok_or_else()` - the linter suggested using `ok_or()` instead.

Tibo-lg:
> @Tibo-lg thanks for all the suggestions and the review! I used your code snippet with `ok_or_else()` - the linter suggested using `ok_or()` instead.

I always forget which one makes the linter happy :)

Also FYI I forgot to mention that I'm actually not able to compile with the current version. Some compile error on rust-bitcoin. It might be good to add a minimal CI script that checks that it builds properly.

yancyribbens:
> I always forget which one makes the linter happy :)

If there are no arguments to the closure passed to `ok_or_else()` then it's better to use `ok_or()` is what I gather.

> Also FYI I forgot to mention that I'm actually not able to compile with the current version. Some compile error on rust-bitcoin. It might be good to add a minimal CI script that checks that it builds properly.

There is a minimal CI script that makes sure it builds: see [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/master/contrib/test.sh#L13).  It sounds to me like you might has some problem locally with cached versions.  Can you build rust-bitcoin?  All cached code is stored in the [Cargo Home](https://doc.rust-lang.org/cargo/guide/cargo-home.html) directory which is normally `~./cargo`.  You can delete this directory and let cargo rebuild it.  You may also need to re-install rustup since rustup lives there too.

Tibo-lg:
> There is a minimal CI script that makes sure it builds: see [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/master/contrib/test.sh#L13).

Ah fair enough I remember that now.

> It sounds to me like you might has some problem locally with cached versions. Can you build rust-bitcoin? All cached code is stored in the [Cargo Home](https://doc.rust-lang.org/cargo/guide/cargo-home.html) directory which is normally ~./cargo. You can delete this directory and let cargo rebuild it. You may also need to re-install rustup since rustup lives there too.

I don't really want to do that, is it because you forced push the branch? Can't you use a commit number instead?

yancyribbens:
> I don't really want to do that, is it because you forced push the branch? Can't you use a commit number instead?

Can you verify rust-bitcoin builds?  I don't think the error is related to this crate.

yancyribbens:
Hmm it looks like this PR is still pointing to my fork so I guess that's not why you're having trouble..  I'm going to change it back to rust-bitcoin since we don't need the fork anymore.

Tibo-lg:
> Hmm it looks like this PR is still pointing to my fork so I guess that's not why you're having trouble.. I'm going to change it back to rust-bitcoin since we don't need the fork anymore.

Yeah it's building now 🎉

yancyribbens:
Unless there's any other feedback I'll probably merge this shortly.

yancyribbens:
Thanks again Murch (http://murch.one) and  Thibaut Le Guilly for the reviews.

PR #27, opened 2023-12-02
Use effective_value in Rust Bitcoin

yancyribbens:
This PR removes the effective_value calculation instead preferring to outsource the calculation to Rust Bitcoin.  Also, temporarily remove the error mod for now since Rust Bitcoin returns Option instead of Result.  Leaving in Draft state until the outcome of https://github.com/rust-bitcoin/rust-bitcoin/pull/2230 is decided.


yancyribbens:
@Tibo-lg maybe you care to comment on this or https://github.com/rust-bitcoin/rust-bitcoin/pull/2230

Tibo-lg:
It doesn't build for me locally:
```
error[E0599]: no method named `effective_value` found for struct `TxOut` in the current scope
  --> src/single_random_draw.rs:46:25
   |
46 |             w_utxo.utxo.effective_value(fee_rate, w_utxo.satisfaction_weight)?;
   |                         ^^^^^^^^^^^^^^^ method not found in `TxOut`
```

yancyribbens:
> It doesn't build for me locally:

Oops, I was asked to make this a stand-alone function instead of a method on TxOut.  I updated this PR so it should work again now.

src/single_random_draw.rs
```diff
@@ -68,22 +43,22 @@ pub fn select_coins_srd<R: rand::Rng + ?Sized>(
...
+        let effective_value: SignedAmount = effective_value(fee_rate, w_utxo.satisfaction_weight, utxo_value)?;
 
-        // skip if effective_value is negative.
-        match effective_value {
-            Some(e) => value += e,
-            None => continue,
-        }
+        value += match effective_value.to_unsigned() {
+            Ok(amt) => amt,
+            Err(_) => Amount::ZERO,
```
Tibo-lg, line 51: Shouldn't you discard the input if the value is negative instead of counting it as zero?

Tibo-lg: Unless I'm missing something, since you're pushing the utxo to the `result` vector anyway it seems like the total value could be lower than what you compute if you add utxos with negative values.

yancyribbens: :facepalm: yes, you're right, this is getting pushed to `result` later on which it shouldn't.  Thanks, I'll update this.

yancyribbens: Ok this has been fixed.  I also expanded the test `select_coins_skip_negative_effective_value`.  Previously, the test had only one utxo which was negative, so it passed.  The test now contains multiple positive effective values and one negative effective value which would then fail if the negative effective value is not discarded.

src/single_random_draw.rs
```diff
@@ -171,32 +145,36 @@ mod tests {
...
+        let mut negative_eff_value = vec![WeightedUtxo {
             satisfaction_weight: Weight::ZERO,
             utxo: TxOut {
                 value: Amount::from_str("1 sat").unwrap(),
                 script_pubkey: ScriptBuf::new(),
             },
         }];
 
-        let result = select_coins_srd(target, FEE_RATE, &mut weighted_utxos, &mut get_rng())
+        weighted_utxos.append(&mut negative_eff_value);
```
Tibo-lg, line 159: why do you create a vector instead of just pushing the additional value?

yancyribbens: Thanks.  I agree it's sloppy to append and we can just use push instead.  Updated now.

yancyribbens:
Merging for now since there seems to be no further feedback.  Waiting on MR https://github.com/rust-bitcoin/rust-bitcoin/pull/2230 to resolve.  Will circle back and remove my fork of rust-bitcoin once resolved.

PR #28, opened 2023-12-17
Feature/iter branch and bound

yancyribbens:
Implement Branch and Bound coin selection.


yancyribbens:
While there is a lot of potential refactors, I think this is ready to move out of draft state.

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
...
+            // The available value of the next iteration.
+            available_value = utxo_candidate_amounts[index + 1..].iter().cloned().checked_sum()?;
+
+            // If the new subtree does not have enough value, we are done searching.
+            if available_value < target {
+                return index_to_utxo_list(best_selection, weighted_utxos);
+            }
+
+            // Start a new selection and add the root of the new subtree to the index selection.
+            index_selection = vec![];
```
Kixunil, line 265: Using `index_selection.clear()` should be faster.

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
+use bitcoin::Amount;
+
+use crate::WeightedUtxo;
+use bitcoin::amount::CheckedSum;
+use bitcoin::blockdata::effective_value;
+use bitcoin::FeeRate;
+
+fn index_to_utxo_list(
+    index_list: Option<Vec<usize>>,
```
Kixunil, line 9: You could accept `Option<&[usize]>` here to allow reusing the allocation in the caller.

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
...
+        //
+        // Check if index_selection is better than the previous known best, and
+        // update best_selection accordingly.
+        else if value >= target {
+            backtrack = true;
+
+            let current_waste = value - target;
+
+            if current_waste <= waste {
+                best_selection = Some(index_selection.clone());
```
Kixunil, line 233:
```rust
match best_selection {
    None => best_selection = Some(index_selection.clone());
    Some(best_selection) => {
        best_selection.clear();
        best_selection.extend_from_slice(&index_selection);
    },
}
```

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
...
+        // * Backtrack to new tree
+        else if backtrack_subtree {
+            // No new subtree left to explore.
+            if index_selection.is_empty() {
+                return index_to_utxo_list(best_selection, weighted_utxos);
+            }
+
+            // Anchor the new subtree at the next available index.
+            // The next iteration, the index will be incremented by one.
+            index = index_selection.remove(0);
```
Kixunil, line 254: This requires shifting the whole thing, can the algorithm be rewritten to use `VecDequeue`?

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
...
+
+            // The available value of the next iteration.
+            available_value = utxo_candidate_amounts[index + 1..].iter().cloned().checked_sum()?;
```
Kixunil, line 257: Doesn't matter much but `copied` would be more clear.

Kixunil (commented):
Looked at obvious performance issues and found some.

src/branch_and_bound.rs
yancyribbens: Nice one!  Looks like that brought the time down to ~`830,564 ns/iter`

yancyribbens: I guess this saves a call to the allocator to allocate a new vec every time and must just mark the current vec as being able to be over-written?  Anyway, surprisingly much faster.

Kixunil: Yes, allocators are slow in general, that's why we try to minimize calling them in Rust.

src/branch_and_bound.rs
yancyribbens: Not sure what you mean by shifting the whole thing.  Although depending on where the pointer is that rust maintains to the vec, maybe it has to do the arithmetic to find the head and remove?  VecDeque should be faster if pop_front() can be used since there's no pointer arithmetic involved.  Will test that idea out.  thanks!

src/branch_and_bound.rs
yancyribbens: I'm not sure this will change performance since since this function isn't really in the "hot path" and only gets called once when the function has finished, unless I'm missing something.  Although it can't hurt either to do this.

src/branch_and_bound.rs
Kixunil: When you call `remove(0)` all contents of `Vec` has to be copied to take its place. Pointer arithmetic is actually infinitely faster and it's what `VecDequeue` uses. `pop_front` is equivalent to `remove(0)` in semantics but much faster.

src/branch_and_bound.rs
Kixunil: Oh, then it may be best to keep it as is.

src/branch_and_bound.rs
yancyribbens: interestingly this seems to have worsened performance slightly.  Note here is what made the compiler happy:

```
            if current_waste <= waste {
                match best_selection {
                    None => best_selection = Some(index_selection.clone()),
                    Some(ref mut best_selection) => {
                        best_selection.clear();
                        best_selection.extend_from_slice(&index_selection);
                    }
                }

                waste = current_waste;
            }

```

I don't think avoiding the reassignment of the stack variable `best_selection` is better here, and based on the benchmarks actually a bit worse.

from about ~830,000 to ~850,000 using the exiting benchmark.

yancyribbens: s/exiting/existing

Kixunil: That's super weird. I know it's not idiomatic but doesn't empty vec work in place of `None`? Maybe the branch is slowing it down.

Another crazy idea: maybe pre-allocate the vectors to the size of the inputs. It wastes quite a chunk of memory but could be much faster and we could trim it at the end (or let the user trim it).

src/branch_and_bound.rs
```diff
@@ -0,0 +1,582 @@
...
+use crate::WeightedUtxo;
+use bitcoin::amount::CheckedSum;
+use bitcoin::blockdata::effective_value;
+use bitcoin::FeeRate;
+
+fn index_to_utxo_list(
+    index_list: Option<Vec<usize>>,
+    weighted_utxos: &mut [WeightedUtxo],
+) -> Option<Vec<WeightedUtxo>> {
+    index_list.map(|i_list| i_list.iter().map(|i: &usize| weighted_utxos[*i].clone()).collect())
```
Kixunil, line 12: I realized this collects the vec which may be unneeded for the callers, so it's better tor return `impl Iterator<Item=WeightedUtxo>` or even better `impl Iterator<Item=&WeightedUtxo>` if we can get the borrow checker to agree.

src/branch_and_bound.rs
yancyribbens: Ah I see, yeah copying the entire vec where the 2nd becomes the first etc would be slow.

src/branch_and_bound.rs
yancyribbens: Yeah I thought of that too, and maybe it's worth doing.  It's possible it doesn't waste any memory of the selection happens to use all inputs but it's unlikely.

src/branch_and_bound.rs
yancyribbens: I tested out a DeQueue in place of a Vec and surprisingly it's far slower.  Went from about 830,000 to 950,000 or so.  Here is the experimental commit:

https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/28/commits/18145ce0c61f1da6d191023f4970a4eb727ad1e7

src/branch_and_bound.rs
yancyribbens: The thing is that it doesn't seem like a fair comparison anymore between the Rust and C++ implementation to pre-allocate and consume more memory for the sake of performance.  I would think the same option would be available in the core C++ implementation and perhaps it's reasoned that the additional memory usage isn't worth the perf increase.

src/branch_and_bound.rs
yancyribbens: Marking as resolved since this seems like a clear win.

yancyribbens:
I found that using primitive types instead of types like `Amount` in the hot path leads to a 2x perf improvement which is crazy.  The benchmark on this branch https://github.com/p2pderivatives/rust-bitcoin-coin-selection/tree/experimental/use-primitive-types runs at ~311,000 ns/iter as opposed to 830,000 ns/iter.  While this version does not yet add a check for high vs low fee like the C++ implementation, even after adding that in, I think this version is much faster than any implementation.

src/branch_and_bound.rs
Kixunil: Looking at Core code I don't see remove there. Maybe if this code resembled it more it'd be faster? I don't understand the code enough to understand the difference (so far I only scanned obvious perf issues).

Kixunil:
That could be because of overflow protection in `Amount` or messed up benchmark. For instance I don't see `black_box` being used in the benchmark so it could be why the results are weird.

yancyribbens:
I did remove a `checked_sum()` in the hotpath (and one that wasn't) so maybe that's in part why.  Also, is there really no `checked_sum()` on [Iter](https://doc.rust-lang.org/std/iter/trait.Iterator.htm)?  Could just manually check each one while summing to see how that stacks up.  Amount just wraps a `u64` so I really didn't expect accessing the `Amount` type to be that much worse.  Will look into `black_box`.  It would be interesting to take a look at the ASM produced also to really understand the performance more - I'm not sure the benchmark is necessarily "weird"..

yancyribbens:
> For instance I don't see black_box being used in the benchmark so it could be why the results are weird.

This article was pretty helpful in understanding blackbox: https://gendignoux.com/blog/2022/01/31/rust-benchmarks.html.  I think the benchmarks are correct however currently without blackbox because here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/28/files#diff-9098d62be93e83524a8371395c973d761a95000d1c295f600a8c808e917c16d9R577  I use the return value which forces the compiler to not optimize away the call to `bnb()`.    If the compiler was optimizing away the call to `bnb()` it would not be measurements of 300,000+ ns per iter and vary based on changes to `bnb`.  The reason for blackbox is to make sure that in the benchmark test itself, the compiler does not optimize away the call to the code you want to benchmark by realizing that since the results aren't being used then don't make the call at all.

yancyribbens:
regarding bench-marking, it looks like criteron would be a better choice.

src/branch_and_bound.rs
yancyribbens: > That's super weird. I know it's not idiomatic but doesn't empty vec work in place of None

As commented in the `/// Returns sections` `None` means there was some type of error such as overflow while empty vec means that everything ran ok, but no match was found.  I'm flexible on the semantics but I think that makes sense as downstream components may want to handle an error different than no results.

src/branch_and_bound.rs
yancyribbens: `copied` is a minor degradation in performance in my testing.  Again, maybe switching to Criterion for better measurements should be a priority, but I show changing this from `cloned` to `copied` reduces performance from 830k ns/iter to 860k ns/iter.

src/branch_and_bound.rs
Kixunil: You can still return `Option` but internally use empty array and only reurn `None` on errors.

src/branch_and_bound.rs
Kixunil: Do I understand correctly that the Core code ignores the first value rather than popping it? So it keeps it in memory? If so, I think that's what makes it faster.

src/branch_and_bound.rs
Kixunil: Something is seriously broken then. There's no good reason why those two should be different. Maybe something for bug report against rustc/llvm if you can minimize it.

src/branch_and_bound.rs
yancyribbens: No, it does pop the value off `curr_selection` using `pop_back()` https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L147.   So in the case where the target is 5, and the solution index [0, 3] was found, the for loop unwinds and index and 3 is popped, then the next iteration occurs, and backtrack happens again, except it unwinds the index back to 0 which is then popped, and the next iteration starts at 1.  So essentially, `pop_back()` is equivalent to the what I've done with `remove(0)`,  I just do it with less iterations.

yancyribbens: Actually, come to think of it, we can just do a memory access here instead of `pop()`.  I tested the change and it reduces the time from 830k ns/iter to 810k ns/iter.  The reason is, we are just doing a clear later.

yancyribbens:
Replaced the `checked_sum()` with `sum()` for the `available_value` calculation in the hotpath since it's not possible to overflow.  This brings the time down to 772 ns/iter.
 
`test branch_and_bound::benches::bench_select_coins_bnb ... bench:     772,163 ns/iter (+/- 393)`

src/branch_and_bound.rs
Kixunil: Popping from back is completely different - and much faster! I really don't understand the code yet (and am very tired right now) so are you doing everything in reverse?

yancyribbens: > Popping from back is completely different - and much faster! I really don't understand the code yet (and am very tired right now) so are you doing everything in reverse?

Yes, however in the case where we are at the last index, the following occurs:

Returning to the example of `[0, 3]`.  At this point, the C++ code does a `pop_back()` so the selection becomes `[0]` then another iteration occurs.  Next iteration, it now does `pop_back()` on `[0]` so we are at an empty list.  This actually isn't very efficient for the edge case where the C++ code does multiple pop_back() functions to backtrack to a new subtree.  An equivalent way to represent this edge case is to store `0` and clear the list instead of 2 `pop_back()` calls.  Finally the other case where we are back tracking one node, I also do a pop back here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/28/files#diff-9098d62be93e83524a8371395c973d761a95000d1c295f600a8c808e917c16d9R24

Kixunil: Well, if you're calling `remove(0)` then you also iterate twice, you just don't see it because the method does it.

yancyribbens: This has been changed to `remove[0]`: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/28/files#diff-9098d62be93e83524a8371395c973d761a95000d1c295f600a8c808e917c16d9R254

yancyribbens: Also, I think you're confusing which iteration I'm talking about.  The way core is doing it, when a backtrack occurs that must back track all the way to the root, it's doing two iterations of the BnB loop.

src/branch_and_bound.rs
yancyribbens: What about:

```rust
fn index_to_utxo_list<'a>(
    index_list: Option<Vec<usize>>,
    weighted_utxos: &'a mut [&'a WeightedUtxo],
) -> Option<impl Iterator<Item=&'a WeightedUtxo>> {
    index_list.map(move |i_list| i_list.into_iter().map(move |i: usize| weighted_utxos[i]))
}
```

and the corresponding test
```rust
    #[test]     
    fn index_to_utxo_list_collect() {
        let mut wu = [&WeightedUtxo {
            satisfaction_weight: Weight::ZERO,
            utxo: TxOut { value: Amount::from_str("1 sat").unwrap(), script_pubkey: ScriptBuf::new() }
        }];
        let i = Some(vec![0]);
        let list: Vec<_> = index_to_utxo_list(i, &mut wu).unwrap().collect();
    }   
```

I think `fn select_coins_bnb(` should use the same return type then `Option<impl Iterator<Item=&'a WeightedUtxo>>`

src/branch_and_bound.rs
yancyribbens: changed to Criterion and it shows less than a 2% variance after making this change (no noticeable change).

src/branch_and_bound.rs
Kixunil: Yeah, I still didn't have time to study the algo...

src/branch_and_bound.rs
Kixunil: Yeah, looks good though I think `&mut [&'a WeightedUtxo]` should be used instead (notice missing lifetime). There shouldn't be the need to force the outer reference into the lifetime, only the inner ones. (But this reaches the limits of my understanding of lifetimes so maybe the borrow checker disagrees.)

src/branch_and_bound.rs
yancyribbens: Understandable.  It takes a bit of study to really understand how the algo works.  I was thinking it would valuable to benchmark the number of utxo combinations that are tried per time interval instead of the number of the time per iteration since sometimes a full iteration is spent just backtracking one or more steps.  So to get a real comparison of how well the algos perform, some better code profiling should be helpful.

Kixunil: I guess smarter people than us already figured that out and just correctly porting the code is the most valuable thing to do right now.

yancyribbens: Yeah it would be nice to have some people who know that part of bitcoin core review this.  I could try an exact copy of the method core uses to backtrack, maybe on a different branch.  Would be interesting to see if there's a difference in timing.

yancyribbens:
Re-opening this as a PR using my remote so people don't get so many notifications on CI failure.

yancyribbens:
New PR open here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30/commits/dc0e5dff34a4fb7bc3a43a6648d8c78dc24b273a

src/branch_and_bound.rs
yancyribbens: @Kixunil I ended up changing the return type to:
) -> Option<std::vec::IntoIter<&'_ WeightedUtxo>> {

here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

The opaque type:  `impl Iterator<Item=&'a WeightedUtxo>` was giving me problems in `lib.rs`.  The compiler won't allow both SRD and BNB to return opaque types.

yancyribbens: > I realized this collects the vec which may be unneeded for the callers, so it's better tor return impl Iterator<Item=WeightedUtxo> or even better impl Iterator<Item=&WeightedUtxo> if we can get the borrow checker to agree.

@Kixunil I am thinking about undoing this change because I think it doesn't make sense anymore.  I'm in the process of updating the interface to return the iteration count https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/76 so it would be a good time to add any other breaking change.  The algorithms no longer uses collect, so I don't think there is any good reason to not simply return a vector instead of an opaque type.  Also returning a concrete type will make it easier to cleanup the return type with an alias.

Kixunil: Unfortunately I no longer remember what this was about.

yancyribbens: > Unfortunately I no longer remember what this was about.

Haha. ok then feel free to disregard.  In short though, initially I was returning a Vector, then switched to return an Iter to remove a collect() that was happening.  That colleect() was refactored out so I think it makes sense to just return a Vector again.  Should have caught the sooner.

PR #30, opened 2024-01-26
Feature/iter branch and bound

yancyribbens:
Implement Branch and Bound search.


src/branch_and_bound.rs
```diff
@@ -1,222 +1,785 @@
...
+        // calculate effective_value and waste for each w_utxo.
+        .map(|wu| (wu.effective_value(fee_rate), wu.waste(fee_rate, long_term_fee_rate), wu))
+        // remove utxos that either had an error in the effective_value or waste calculation.
+        .filter(|(eff_val, waste, _)| eff_val.is_some() && waste.is_some())
+        // unwrap the option type since we know they are not None (see previous step).
+        .map(|(eff_val, waste, wu)| (eff_val.unwrap(), waste.unwrap(), wu))
+        // filter out all effective_values that are negative.
+        .filter(|(eff_val, _, _)| eff_val.is_positive())
+        // all utxo effective_values are now positive (see previous step) - cast to unsigned.
+        .map(|(eff_val, waste, wu)| (eff_val.to_unsigned().unwrap(), waste, wu))
```
murchandamus, line 162: I don’t understand why you would change the type. What’s the advantage of casting to unsigned here?

src/branch_and_bound.rs
```diff
@@ -1,222 +1,785 @@
...
+            }
 
-    let lower_bound = target;
-    let upper_bound = cost_of_change + lower_bound;
+            // Anchor the new subtree at the next available index.
+            //
+            // if our index selection is: [0,1,2,3]
+            // then copy the head 0 into index.
+            // At the end of this loop, index is incremented to 1.
+            // Therefore, we will be starting our next search tree at index 1.
```
murchandamus, line 288: It sounds to me like you are using a different order of exploration than Bitcoin Core’s BnB implementation. Bitcoin Core uses a depth-first search, the traversed UTXO combinations of four UTXOs 1, 2, 3, and 4 would be (underscore is an omitted UTXO):

```
{} {1} {12} {123} {1234} {123_} {12_} {12_4} {12__} {1_} {1_3} {1_34} {1_3_} {1__} {1__4} {1___} {_} {_2} {_23} {_234} {_23_} {_2_} {_2_4} {_2__} {__} {__3} {__34} {__3} {___} {___4} {____}
```

I’ve described the same search pattern in a lot of detail here on the CoinGrinder PR: https://github.com/bitcoin/bitcoin/pull/27877/commits/6cc9a46cd0f4ed80d4523bbef1508142e0c80d27

I’m not particularly familiar with Rust, and it’s not obvious to me whether the search strategy employed here will enumerate all possible combinations. In case you want to verify that, I would suggest that you print each evaluated input combination your search traverses to check whether all input combinations that should be enumerated get traversed. I suspect that immediately jumping to the next index as described in the comment here would skip part of your search tree.

murchandamus (commented):
Hey Yancy, I have no familiarity with Rust, but it seems to me that you may be implementing a different search order than the one we use in Bitcoin Core, and it seems to me that the way the backtracking is implemented here may skip part of the search tree.

At this point, I’ve provided a number of explanations on how BnB works. It seems to me that the way I think about the algorithm and explain it doesn’t click for you. Perhaps my [documentation of CoinGrinder](https://github.com/bitcoin/bitcoin/pull/27877/commits/6cc9a46cd0f4ed80d4523bbef1508142e0c80d27) is useful to you, but I’m afraid that I will not be able to allocate additional time to this project.

src/branch_and_bound.rs
```diff
@@ -1,222 +1,785 @@
...
+/// binary tree with a maximum depth n where n is the size of the UTXO pool to be searched.
+///
+/// See also core: <https://github.com/bitcoin/bitcoin/blob/f3bc1a72825fe2b51f4bc20e004cef464f05b965/src/wallet/coinselection.cpp>
+///
+/// Returns a vector of `WeightedUtxo` that meet or exceed the target `Amount` when summed.
+/// The `Amount` returned will not exceed the target by more than target + delta where delta is
+/// the cost of producing a change output.
+///
+/// The vector returned seeks to minimize the waste, which is the difference between the target
+/// `Amount` and sum of the vector returned.  If no match can be found, None is returned.
```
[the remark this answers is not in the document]
yancyribbens: Good catch, Thanks!

src/branch_and_bound.rs
```diff
@@ -1,222 +1,785 @@
...
+// only one root node that is less than the total, so the next iteration occurs.  The second
+// iteration examines a tree where 4 is the root and the left branch is 3.
+//      o
+//     /
+//    4
+//   /
+//  3
+//
+// At this point, the total sums to 7 which exceeds the target of 5.  7 may be recorded as a
+// solution with a waste of 2 (7 - 5). Next, 3 is removed from the left branch and it becomes
```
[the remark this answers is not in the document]
yancyribbens: Good catch, Thanks!

src/branch_and_bound.rs
yancyribbens: It may be possible to leave this as signed.  Since it's a positive value, I am not sure if there's a reason to cast this to Signed, although I see your point that maybe the computation can carry through with unsigned.  Will investigate further.

yancyribbens:
> Hey Yancy, I have no familiarity with Rust, but it seems to me that you may be implementing a different search order than the one we use in Bitcoin Core, and it seems to me that the way the backtracking is implemented here may skip part of the search tree.

I'll double check to make sure I'm not missing any solutions.  That would be embarrassing if the backtracking does indeed skip part of the search tree.  I'm fairly certain that this is the same backtracking pattern as bitcoin-core, and I think you are one of the few people I could ask for review.

> At this point, I’ve provided a number of explanations on how BnB works. It seems to me that the way I think about the algorithm and explain it doesn’t click for you. Perhaps my [documentation of CoinGrinder](https://github.com/bitcoin/bitcoin/pull/27877/commits/6cc9a46cd0f4ed80d4523bbef1508142e0c80d27) is useful to you, but I’m afraid that I will not be able to allocate additional time to this project.

Thanks for the tip to checkout Coin Grinder, any additional resources are very welcome.  Also thanks again for your review.  I'm sad to hear you can't allocate any more time to this project.  I really appreciate the time you were able to allocate.  It's a particularly difficult part of bitcoin-core to understand and the number people that can review this work is next to zero, which makes a difficult area to work on.  Your reviews will be sorely missed.  Of course if you reconsider in the future I'd be more than thrilled.

src/branch_and_bound.rs
```diff
@@ -1,222 +1,785 @@
...
+//   / \
+//      3
+//
+// We next try adding 2 to the inclusion branch.
+//      o
+//     /
+//    4
+//   / \
+//  2   3
+//
```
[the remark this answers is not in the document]
yancyribbens: Thanks, I've updated the diagrams.  You're right, it was confusing that I had the lower depths in some diagrams at higher point in the tree.  Also thanks for the full diagram with the negation symbol to show what's negated and what's not.  I suppose that can help if one doesn't realize the right branch is always negated and the left branch is always included.

src/branch_and_bound.rs
yancyribbens: As with bitcoin core bnb, the first thing that happens is a reverse sort: https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L114.   four UTXO combinations 1,2,3,4 would be searched depth first starting at 4.

`{} {4} {43} {432} {4321}`

If a print statement is add here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/feature/iter-branch-and-bound/src/branch_and_bound.rs#L318 where the next depth is explored (depth + 1).

`println!("{}", eff_value)`

The results will be 4,3,2,1 if the test for 10 is run with the following:

`cargo test select_coins_bnb_ten -- --nocapture`

Note that `-- --nocapture` is needed when running the test suit to print to stdout.

I'm struggling to find other samples that would prove this is a depth first search.  In this case, the search stops after 4,3,2,1 because once 4 is remove from the inclusion branch, there is no longer enough value left to reach 10, so the search ends.

> I’m not particularly familiar with Rust

If there's anything about rust that is making your review slow or not possible I'm happy to help answer any questions.

src/branch_and_bound.rs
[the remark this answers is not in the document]
yancyribbens: fixed

src/branch_and_bound.rs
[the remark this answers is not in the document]
yancyribbens: fixed

src/branch_and_bound.rs
yancyribbens: One more less trival example five to show this matches what bitcoin-core would find.

Placing a print statement at L318, will print:
`4,3,2,1,3,2`

that is:
`{} {4} {4,3} {4,2} {4,1} {3} {3,2}`

This is the same DFS and set evaluation that core performs and does not skip any part of the search tree.

yancyribbens: run `cargo test select_coins_bnb_five -- --nocapture` to duplicate

murchandamus: Given UTXOs with the effective values { 10, 7.00000005, 6.00000005, 6, 3, 2, 1.00000005 } and a target of 18, what is the best solution and what does your implementation return?

yancyribbens: I added a test (in the quickest possible way for now) to this branch https://github.com/yancyribbens/rust-bitcoin-coin-selection/tree/test/add-extended-tests.

see commit: https://github.com/yancyribbens/rust-bitcoin-coin-selection/commit/b46e8bd1e7ee4f88ad80d0834bedd580870de416

Using 50 sats for a cost_of_change:
the implementation returns: `6.00000005, 6, 3, 2, 1.00000005`

Happy to flush out more tests in detail, but just wanted to reply with this to show that the implementation does work..

yancyribbens: oops sorry, see this commit https://github.com/yancyribbens/rust-bitcoin-coin-selection/commit/fcd1908295821423839254e7e596c55bb263f53c

yancyribbens: On further inspection, and drawing out a number of binary tree diagrams, I do think this misses some solutions because of the way I implemented backtracking with backtrack subtree.  That's embarrassing but hopefully fixing the backtrack won't be too painful.  I'll take a look at Coin Grinder also for some inspiration.

yancyribbens: >  what is the best solution and what does your implementation return?

I think the best solution is: [7.00000005, 6, 3, 2] = 18.00000005

My implementation returns [6.00000005, 6, 3, 2, 1.00000005] = 18.0000001

Thanks for providing a better this test case.  I see now that once at a leaf node, there is another backtrack case where we don't backtrack once and we don't backtrack all the way back and omit the root, but another case in between the two.

yancyribbens: What I meant is, there is another backtrack case where we don't backtrack once and we don't backtrack all the way back and omit the root, but another case in between the two.

murchandamus: The best solution is `{10, 6, 2}`.

I feel like I should explain why I told you that I don’t have time for your project. I am frustrated because I feel that I have provided numerous resources to help understand the algorithm, and am kept repeating myself at this point. You should address this project’s issues at the root by implementing adequate test coverage and taking the time to _fully understand_ what you are trying to implement.

My experience is that when I try to point you in a direction that should lead to you _discovering_ and _understanding_ what the problem is, it feels to me like you are putting in too little effort to understanding my feedback and instead of pondering why I write what I suggest and following my advice, arrive at uncharitable interpretations like I don’t understand something as obvious as your UTXOs being sorted by descending effective value and immediately ask for more help. So, on the one hand, you are asking me to invest my time, guilt-trip me by saying that you don’t have any other reviewer, but on the other hand, you make me feel that you seem to think that you already know better than me anyway.

> Thanks for providing a better this test case. I see now that once at a leaf node, there is another backtrack case where we don't backtrack once and we don't backtrack all the way back and omit the root, but another case in between the two.

As I [wrote last week](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30#discussion_r1500685154): 

> Following this advice would help you understand why I don’t use something like backtrack_subtree, but instead [backtrack twice](https://github.com/bitcoin/bitcoin/blob/1ac627c485a43e50a9a49baddce186ee3ad4daad/src/wallet/coinselection.cpp#L402-L411). I would expect that it would be helpful to investigate what issues are covered by the BnB tests in Bitcoin Core. You may want to replicate more of those tests. To that end, it may be useful to look at the old tests and my reimplementation: [diff](https://github.com/bitcoin/bitcoin/compare/e6dbf48c64a3c267425911d768a77b95a6c3d078..b27e06608fe9f58a2a6377547738758a7d0bbe7c), [list of commits in the reimplementation](https://github.com/bitcoin/bitcoin/pull/28985/commits)

You have to backtrack twice. I.e. if your selection out of the UTXOs {ABCDEFG} was {ABDEF} you need to deselect the last {ABDEF} ⇒ {ABDE}, then move the remaining last selected one down: {ABDE} ⇒ {ABDF}. This is explained at least twice in the CoinGrinder documentation, which is why I linked you to it, and should become apparent if you think about the order of all leaves in a depth-first search and play through which ones you can skip but which ones you have to visit e.g. on a whiteboard. 

I really appreciate your interest in coin selection and how long you have been sticking with this project, but I  would appreciate if I saw more proactive effort in comprehending my feedback instead of making me repeat myself, because it is making me feel that you do not respect my time.

murchandamus: By the way, in case that was not clear: CoinGrinder is a reimplementation of _exactly_ the same algorithm as BnB except that there is neither excess, nor an upper bound for the target, because CoinGrinder creates a change output.

yancyribbens: My intent was not to guilt trip you into reviewing.  Apologies if that's how I came off.  It's just that in my opinion, you wrote the book on coin-selection and are the expert which is why I seek your input and review.  I feel like this implementation is much better than my first attempt, but as you pointed out still anemic in how the backtrack is currently working.  I tried to get others to look at it and review it, but none could really follow the logic in bitcoin core well enough to give me feedback on the most critical component.  To be fair I made a blunder by trying to optimize what I thought was the behavior instead of more dutifully trying to copy the semantics which I take responsibility for.  Anyway, thanks as for providing feedback and if there is a future for working together, I will try to more diligently ponder your advice, feedback and provided resources before responding with questions/assumptions.  Also, thanks for explaining more about your frustration and where you are coming from.  You saved me making another mistake with this implementation by reviewing this before it was merged.  I'm not sure how to repay you but consider me in your debt.

murchandamus: Thank you for your understanding. Yes, I do note that this implementation is much better than the first.

In my experience, trying to optimize early or taking shortcuts tends to break things in ways that are hard to diagnose more often than not. While it may feel slower while you are doing it, I find that I’m overall much faster when I take the time to first make the simplest possible implementation, ensure that it is correct, and only then consider whether it needs to be made more efficient. If it needs to be improved, I do one improvement at a time, so I know what I need to test and what change broke the correctness if any test fails.

Some things that helped me over time:

- Commit early, commit often
- Slow is smooth, smooth is fast
- Assume that every untested aspect of your implementation is broken

You can probably make your tests much easier to extend and maintain by creating some helper methods, check out for example `TestBnBSuccess` here: https://github.com/bitcoin/bitcoin/pull/28985/commits/4a95b234918e9f714ac6f707defe17a777225e47#diff-36088c93368e137d955348aba223985bd4f198f2aaecd626c830f4612ca884c8R100-R133

yancyribbens:
@murchandamus I've removed the optimization backtrack_subtree.  Thanks for the additional test case that help show the problem with this implementation.  The current implementation finds the correct `{ 10, 6, 2 }` solution.  Also, I haven't forgotten about the other additions to the test suit you recommended; I'll work on adding more coverage to this PR.  At any rate, I think the core implementation now matches more precisely what bitcoin cores implementation does and ought to find the same solutions.

yancyribbens:
@murchandamus I looked over your checklist of additional tests to add as well as the questions you posed.  I think your question of what will cause BnB to fail was the most interesting..  Please see bellow.

> What happens if you have multiple UTXOs that have only slightly different values?

`select_coins_bnb_set_size_five()` and select_coins_bnb_set_size_seven() both have slightly different values

> What happens if you have many UTXOs?

`select_coins_bnb_set_size_seven()` has many UTXOs.

> What happens if there are many alternative solutions that are not equivalent, when the first is not the best?

This is exemplified by test `more_inputs_when_cheap`.  The test sets the current fee_rate to be 10 sat per kwu and the long term fee rate to be 20 sat per kwu.  Therefore, the more inputs that are selected the better.  A target of 6 is created and a utxo pool [1, 2, 3, 4].  The search routine first finds solution [2, 4] which is not the best solution.  Since it is cheaper now (long_term_fee is higher) each utxo that is added reduces the waste.  Therefor, the next solution that is found [1, 2, 3] has a lower waste metric and is found second. 

> Instead of testing with a feerate of 0, make a helper function that allows you to create UTXOs per their effective value, then run your tests at high and low feerates. Adjust the expected results accordingly.

`select_coins_bnb_consume_more_inputs_when_cheap()` and `select_coins_bnb_consume_less_inputs_when_expensive()` both test the behavior with different combinations of long_term and short_term feerate.  Not to ignore your suggestion about a testing helper function that allows the creating of UTXOs at different effective values, although I was thinking of restructuring BnB so that it takes a list of effective_values instead of a list of UTXOs.  That way, in lib, we would compute the effective_value once and pass the results to either BnB, SRD or whatever other search function is needed.  However, I wasn't planning to do that as part of this PR.  I don't know how that might change your suggestion to add a helper that calculates effective_value.

> Add tests in which BnB fails to find a solution or fails to find the best solution. When does this happen and why?

`select_coins_bnb_exhaust` fails to find a solution because the iteration limit occurs before a solution is found.  This happens because the target is created in such a way that permutations tested first fail and that the correct permutation does not occur until later on in the search routine.  For an intuition of why some permutations occur later than others, consider the most simple example where we have a tree of two, a left branch and a right branch.  Naturally, the algorithm will always select the left branch first, so, the solution that uses the right branch takes the most iterations and the solution with only the left branch takes the least.  Let's call the left branch 1 and the right branch 0.  Another example with numbers from 1 to 10.  The number that takes the most iterations to find is 1, and the number that takes the least iterations to find is 4.  That's because 4 only needs to go down the left branch to find a solution.  Contrast this with 1, which must try the left branch before backtracking and going down the right branch again and again creating a branching pattern 1010101.  Therefore setting the iteration limit to 5 will cause BnB to not find a solution for 1 given a set 1 - 10.

make_hard in bitcoin core which causes the algorithm to reach the iteration limit 100,000 works along this principle.  Call set A 1111 and set B 0000, then the target is set to be the sum of A 1111.  Set C is created by interweaving Set A with Set B creating a set 01010101.  Set C is used as the utxo_pool and the target is the sum of set A.  Naturally 1 is tried first before 0 (the left branch is tried first before the right). It follows that since the left branch 1 is tried first, then other permutations of 1 will be computed before 0. That is, 11 101 111 110 will be tried before 01 etc.

> Test what happens when you run into the iteration limit (i.e. you should have cases in which you find no solution, not the best solution)

see select_coins_bnb_exhaust and select_coins_bnb_exhaust_v2

> Optionally make a test that generates a random UTXO pool and target=
 where you compare your outcome with the result of a:

I still have this as a TODO.  I'm not sure if I'll include it as part of this PR or a followup.

yancyribbens:
Fixed a few more typos in the comments, no code changes.

murchandamus:
Starting my review

murchandamus (commented):
Much better!

> @murchandamus I looked over your checklist of additional tests to add as well as the questions you posed. I think your question of what will cause BnB to fail was the most interesting.. Please see bellow.
> 
> > What happens if you have multiple UTXOs that have only slightly different values?
> 
> `select_coins_bnb_set_size_five()` and select_coins_bnb_set_size_seven() both have slightly different values

What I meant to point out here was that if you have a UTXO pool that has many UTXOs with slightly different effective values, e.g.: the 200 UTXOs `{0.00100000, 0.00100001, 0.00100002, 0.00100003, …, 0.00100198, 0.00100199}`, you can quickly exhaust your iteration limit, if a combination of multiple of these is close to the target (here for example 0.05). But you got that covered now with the iteration limit tests!

> > What happens if you have many UTXOs?
> 
> `select_coins_bnb_set_size_seven()` has many UTXOs.

Regular end-user wallets easily venture into the territory of dozens of UTXOs, especially since DCA for small amounts has become so popular. Enterprise wallets can easily accumulate thousands of UTXOs. The biggest wallet I’ve seen had over 550'000 UTXOs. When I recommend testing a wallet with many UTXOs I was thinking of something like ~1000 UTXOs, not 7. In a follow-up PR, you could perhaps think about how you may explore a search tree more quickly that has many duplicates. 

> > What happens if there are many alternative solutions that are not equivalent, when the first is not the best?
> 
> This is exemplified by test `more_inputs_when_cheap`. The test sets the current fee_rate to be 10 sat per kwu and the long term fee rate to be 20 sat per kwu. Therefore, the more inputs that are selected the better. A target of 6 is created and a utxo pool [1, 2, 3, 4]. The search routine first finds solution [2, 4] which is not the best solution. Since it is cheaper now (long_term_fee is higher) each utxo that is added reduces the waste. Therefor, the next solution that is found [1, 2, 3] has a lower waste metric and is found second.

Yup, good! This is also covered with `select_coins_bnb_set_size_seven()` now.

> > Instead of testing with a feerate of 0, make a helper function that allows you to create UTXOs per their effective value, then run your tests at high and low feerates. Adjust the expected results accordingly.
> 
> `select_coins_bnb_consume_more_inputs_when_cheap()` and `select_coins_bnb_consume_less_inputs_when_expensive()` both test the behavior with different combinations of long_term and short_term feerate. Not to ignore your suggestion about a testing helper function that allows the creating of UTXOs at different effective values, although I was thinking of restructuring BnB so that it takes a list of effective_values instead of a list of UTXOs. That way, in lib, we would compute the effective_value once and pass the results to either BnB, SRD or whatever other search function is needed. However, I wasn't planning to do that as part of this PR. I don't know how that might change your suggestion to add a helper that calculates effective_value.

That’s okay, you could do it in a follow-up. My general idea would be that your UTXOs should have a weight, and when you pass a feerate to your function, it automatically calculates the corresponding fee. That way it would be easy to create a UTXO that has a given effective value, or a UTXO that has a given amount but still a fee.

> > Add tests in which BnB fails to find a solution or fails to find the best solution. When does this happen and why?
> 
> `select_coins_bnb_exhaust` fails to find a solution because the iteration limit occurs before a solution is found. //…snip…//

Looks good!

> > Test what happens when you run into the iteration limit (i.e. you should have cases in which you find no solution, not the best solution)
> 
> see select_coins_bnb_exhaust and select_coins_bnb_exhaust_v2

Yeah, they both do not find a solution, but what happens if you find a solution before running into the iteration limit, but you are not done searching?

> > Optionally make a test that generates a random UTXO pool and target=
> > where you compare your outcome with the result of a:
> 
> I still have this as a TODO. I'm not sure if I'll include it as part of this PR or a followup.

Sure that’s not the easiest. You could perhaps do something like [this Knapsack test I wrote](https://github.com/bitcoin/bitcoin/pull/28985/commits/cd6e3e59c07f490c3b6a7b86a330c16ad9b880fa#diff-36088c93368e137d955348aba223985bd4f198f2aaecd626c830f4612ca884c8R279-R318) to generate a diverse UTXO set and then randomly pick seven UTXOs, sum them up, and use that as target for a BnB search. The test would then assert that you should find a solution, and that this solution has a waste that is equal or lower than if you had used the input set you randomly picked.

I would also suggest that you write a test in which a solution with excess and fewer inputs is preferred over a solution with more inputs at high feerates.

src/branch_and_bound.rs
```diff
@@ -1,222 +1,859 @@
...
+            FeeRate::ZERO,
+            FeeRate::ZERO,
+            &mut weighted_utxos,
+        )
+        .unwrap()
+        .collect();
+
+        assert_eq!(list.len(), 2);
+        assert_eq!(list[0].utxo.value, Amount::from_str("3 cBTC").unwrap());
+        assert_eq!(list[1].utxo.value, Amount::from_str("1 cBTC").unwrap());
```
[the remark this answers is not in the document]
murchandamus: Ah nevermind that, you mentioned you were looking to do that in a follow-up.

src/branch_and_bound.rs
```diff
@@ -1,222 +1,859 @@
...
+// Besides the difference between amount and target, current_waste stores the difference between
+// utxo fee and utxo_long_term_fee.
+//
+// If the iteration adds a new node to the inclusion branch, besides incrementing the accumulated
+// value for the node, the waste is also added to the current_waste.  Note that unlike value,
+// waste can actually be negative.  This happens if there is a low fee environment such that
+// fee is less than long_term_fee.  Therefore, the only case where a solution becomes more
+// wasteful, and we may bound our search because a better waste score is no longer possible is:
+//
+//  1) We have already found a value that matchs the target and the next solution is greater then
```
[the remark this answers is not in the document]
yancyribbens: Good question.  Added a fixup commit to make this more clear hopefully: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30/commits/3a402e4109e5b21ef46c5370ac117d14b5fad8fc

src/branch_and_bound.rs
```diff
@@ -1,222 +1,859 @@
...
+//        4
+//       /
+//      3
+//     /
+//    2
+//
+// We now stop our search again noticing that 3 and 2 equals our target as 5, and since this
+// solution was found last, [3, 2] overwrites the previously found solution [4, 1].  We next
+// backtrack and exclude our root node of this tree 3.  Since our new sub tree starting at 2
+// doesn't have enough value left to meet the target, we conclude our search at [3, 2].
```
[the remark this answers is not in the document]
yancyribbens: Thanks, I expanded the explanation here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30/commits/a5cbf1f37caae01853850677ca578e63a492eb9a

src/branch_and_bound.rs
```diff
@@ -1,222 +1,859 @@
...
+        assert_eq!(list[2].utxo.value, Amount::from_str("2 cBTC").unwrap());
     }
 
     #[test]
-    fn find_solution_5_btc() {
-        let five_btc = FOUR_BTC + ONE_BTC;
-        let utxo_match = find_solution(five_btc, COST_OF_CHANGE, &mut UTXO_POOL.clone()).unwrap();
-        let expected_bool_vec = vec![false, true, true, false];
-        assert_eq!(expected_bool_vec, utxo_match);
+    fn select_coins_bnb_ten() {
```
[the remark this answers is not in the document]
yancyribbens: Good idea.  Will add this into the next PR :+1:

yancyribbens:
@murchandamus 
> I would also suggest that you write a test in which a solution with excess and fewer inputs is preferred over a solution with more inputs at high feerates.

Good idea.  I added a test here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30/commits/92c4e6b9424a9546339d03abb13e1e0238a7303e

> Yeah, they both do not find a solution, but what happens if you find a solution before running into the iteration limit, but you are not done searching?

Good question.  I added a test to show what happens here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30/commits/0114f2a2f41f1eb48bc68555b4f574041c21c98d

I've made note for the next PR along with some other refactors to
- use reasonable fee_rate and effective value
- consolidate common tests with a helper function
- keep at least on test with a zero feerate.
- optionally make a test that generates a random UTXO pool and target

I think the most critical comments have been addressed.  If the fixups all look good I'll squash and merge.  Thanks for all the feedback!

yancyribbens:
As a side note, it would be nice maybe to return the iterations in some way which would make all of the iteration testing more meaningful.

src/branch_and_bound.rs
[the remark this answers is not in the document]
murchandamus: Looks good to me

src/branch_and_bound.rs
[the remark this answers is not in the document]
murchandamus: Good, thanks

src/branch_and_bound.rs
```diff
@@ -0,0 +1,861 @@
...
+        let vals = (0..15).enumerate().flat_map(|(i, _)| {
+            let a = base.pow(15 + i as u32) as u64;
+            target += a;
+            vec![a, a + 2]
+        });
+
+        let mut vals: Vec<_> = vals.map(Amount::from_sat).collect();
+
+        // Add a value that will match the target before iteration exhaustion occurs.
+        vals.push(Amount::from_sat(target));
```
murchandamus, line 847: Looks good, nice changes. 

One nit:
Regarding the `select_coins_bnb_exhaust_with_result()` test added in https://github.com/p2pderivatives/rust-bitcoin-coin-selection/commit/0114f2a2f41f1eb48bc68555b4f574041c21c98d:
Sorry, I expressed myself imprecisely. I meant that you should consider a test that finds _a solution_ but not the best solution due to hitting the iteration limit first. You have a somewhat complicated setup in this test with adding the powers-of-2 and powers-of-2+2 sats UTXOs, and I was first scratching my head a bit how you would even find any solution because your `cost_of_change` is set to 1, and therefore using any of the "+2 UTXOs" would mean that you do not find the solution. Only then did I see that you added another UTXO that exactly matches the target. That means of course that BnB finds the best solution in the first iteration and then simply exhausts the iteration limit afterwards.

I guess that works, but it also means that the exact values you add in the loop basically don’t matter and the test is unnecessarily hard to review.

murchandamus (commented):
The algorithm looks good to me now and the test coverage seems to be fairly complete.

What I haven’t looked into, and remains an open question to me is how the algorithm integrates with the greater code base. The tests set magic numbers for `feerate`, `long term feerate estimate`, `fees`, and `cost_of_change`. How is `select_coins_bnb` called from the transaction building? How do the UTXOs get preprocessed? How is the algorithm integrated into the transaction sending, and how do you ensure that the UTXOs are passed with correct and complete values for the call? When transaction building gets back an input set that does not create change, will it build the transaction correctly, skip the change output and drop the excess to the fees?

I’m not sure whether that’s in scope for this PR, but you may want to scan the PR one more time with the interface in mind, and perhaps consider adding some form of integration test that includes having a wallet set up with UTXOs and a transaction being built from there at a specific feerate that results in a changeless transaction that used BnB under the hood.

murchandamus:
> As a side note, it would be nice maybe to return the iterations in some way which would make all of the iteration testing more meaningful.

In CoinGrinder I return the iteration count in the selection result object for exactly that reason.

yancyribbens:
> Enterprise wallets can easily accumulate thousands of UTXOs. The biggest wallet I’ve seen had over 550'000 UTXOs. When I recommend testing a wallet with many UTXOs I was thinking of something like ~1000

A few more TODOs I forgot to mention earlier for the next iteration:

- Add testing for much larger sets.
- In a follow-up PR, you could perhaps think about how you may explore a search tree more quickly that has many duplicates.

yancyribbens:
> The algorithm looks good to me now and the test coverage seems to be fairly complete.

Thanks!

> What I haven’t looked into, and remains an open question to me is how the algorithm integrates with the greater code base. The tests set magic numbers for feerate, long term feerate estimate, fees, and cost_of_change. How is select_coins_bnb called from the transaction building? How do the UTXOs get preprocessed? How is the algorithm integrated into the transaction sending, and how do you ensure that the UTXOs are passed with correct and complete values for the call? When transaction building gets back an input set that does not create change, will it build the transaction correctly, skip the change output and drop the excess to the fees?

I've been looking into this, although I'm still exploring these questions.  @Tibo-lg can speak more to this than I can, but Rust-DLC has some simple wallet interface that combines coin-selection with electrs to create wallet functionality of some fashion.   My intention with the project is to include this as a rust-bitcoin crate to make it available to anyway as a basic building block for wallet construction.  I've been meaning to try out electrs and interface that with some simple wallet software as well.  Anyway, I'm not sure all the mechanics of RPC to facilitate transaction creation/sending etc yet, although I'm looking forward to exploring that more next :)

> I’m not sure whether that’s in scope for this PR, but you may want to scan the PR one more time with the interface in mind, and perhaps consider adding some form of integration test that includes having a wallet set up with UTXOs and a transaction being built from there at a specific feerate that results in a changeless transaction that used BnB under the hood.

I'm not satisfied with the API yet enough to mark this as complete and ready for use, although I'm satisfied enough to merge this PR shortly.  There's still some ongoing discussion about the interface and getting a review to bring this into rust-bitcoin.  Also, there's still work to be done to add fuzz testing which I think is important to make sure it's a high quality implementation.

src/branch_and_bound.rs
murchandamus: What you have already covers the main concern: I was looking to see a test confirm that the selection function will return a result even if it does not manage to conclude the search. I just mean that if you’ll have a single UTXO that immediately provides the best solution, the remainder of the test could be much simpler.

murchandamus:
> I'm not satisfied with the API yet enough to mark this as complete and ready for use, although I'm satisfied enough to merge this PR shortly. There's still some ongoing discussion about the interface and getting a review to bring this into rust-bitcoin. Also, there's still work to be done to add fuzz testing which I think is important to make sure it's a high quality implementation.

If you are going to make a crate that is called by completely unrelated software, you would probably want to put more verification of your assumptions into your code. E.g. you may want to ensure that every UTXO has a reasonable weight and/or fee beside the amount, if fee and weight are both present that they match with the feerate, etc.

yancyribbens:
> If you are going to make a crate that is called by completely unrelated software, you would probably want to put more verification of your assumptions into your code. E.g. you may want to ensure that every UTXO has a reasonable weight and/or fee beside the amount, if fee and weight are both present that they match with the feerate, etc.

I'm not sure if you've seen [InputWeightPredict](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockda/transaction.rs#L1494) in Rust Bitcoin.  I was thinking of next changing the API to accept a list of WeightPredictions as well as values.  That way, every UTXO can have an associated weight prediction instead of requiring the consumer to calculate the satisfaction weight..

Issue #31, opened 2024-02-06
Change all return types to iter


yancyribbens:
In case consumers don't need a Vec, an iterator type can be returned, then collected() for the cases where a vec is needed.

Change return type pf `select_coins_srd()`, `select_coins()` and `select_coins_bnb()` to:

```
-> Option<impl Iterator<Item=&'a WeightedUtxo>> 
```

This should ideally be done at the same time.

yancyribbens:
closed by: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

yancyribbens:
closed by https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

PR #33, opened 2024-02-13
Return Iterator instead of Vector for SRD

yancyribbens:
Return Iterator instead of Vector for SRD.

The caller may prefer to work with an Iterator instead of a vec.  If the caller prefers a vec, they can use `collect()`.  This is the first commit to close https://github.com/p2pderivatives/rust-bitcoin-coin-selection/issues/31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,15 +59,11 @@ pub struct WeightedUtxo {
 #[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
 pub fn select_coins<T: Utxo>(
     target: Amount,
-    cost_of_change: u64,
+    _cost_of_change: u64,
     fee_rate: FeeRate,
     weighted_utxos: &mut [WeightedUtxo],
-    utxo_pool: &mut [T],
-) -> Option<Vec<WeightedUtxo>> {
-    match select_coins_bnb(target.to_sat(), cost_of_change, utxo_pool) {
-        Some(_res) => Some(Vec::new()),
-        None => select_coins_srd(target, fee_rate, weighted_utxos, &mut thread_rng()),
-    }
+) -> Option<impl Iterator<Item = &'_ WeightedUtxo>> {
+    select_coins_srd(target, fee_rate, weighted_utxos, &mut thread_rng())
 }
 
 /// Select coins using BnB algorithm similar to what is done in bitcoin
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -59,15 +59,11 @@ pub struct WeightedUtxo {
 #[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
 pub fn select_coins<T: Utxo>(
     target: Amount,
-    cost_of_change: u64,
+    _cost_of_change: u64,
```
Kixunil, line 62: What? Why is this not used?

yancyribbens: It's only used by bnb, not srd and frankly the current version of bnb that's in master is broken so I just commented it out.  I'm still waiting to get some review on https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30, especially those that created issues.

Kixunil: Weird that it doesn't take into account the cost of change (why would anyone want such algo then?) anyway, just remove the argument?

yancyribbens:
I should have marked this draft since it's not really intended to merge,
at least not until the bnb pr merges.

> Weird that it doesn't take into account the cost of change (why would anyone want such algo then?) anyway, just remove the argument?

As in bitcoin core, bnb uses cost of change to determine if an exact
match can be found where the match is not greater than the target + cost
of creating a change output (cost_of_change).  However, if bnb can't
find an exact match the fallback strategy is to use SRD (single random
draw) to choose any selection which will surely create a change output,
so the cost_of_change isn't needed in the fallback scenario.

yancyribbens:
Nope, looks like I was right the first time.  CHANGE_LOWER is used to
avoid excessively small amounts of change, so the target is increased by
that amount.  Although, it looks like a bug was found in core since I
last looked where if the cost_of_change is greater than CHANGE_LOWER
then there's not enough to pay for creating the change output and so
none is created.  That needs to be added in this crate as well.

yancyribbens:
Anyway, I'm primarily trying to decide on an interface to make "stable".
It's either:

1)
```
pub fn select_coins_srd<'a, R: rand::Rng + ?Sized>(
    target: Amount,
    fee_rate: FeeRate,
    weighted_utxos: &'a mut [WeightedUtxo],
    rng: &mut R,
) -> Option<impl Iterator<Item = &'a WeightedUtxo>> {
```
or 2)
```
pub fn select_coins_srd<'a, R: rand::Rng + ?Sized>(
    target: Amount,
    fee_rate: FeeRate,
    weighted_utxos: &'a mut [&'a WeightedUtxo],
    rng: &mut R,
) -> Option<impl Iterator<Item = &'a WeightedUtxo>> {
```

I think the second actually is the most flexible for the consumer
although I'm not sure.  Either would probably be fine, just don't
want to need to change it in the future.

yancyribbens:
moved changes to https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30

PR #37, opened 2024-02-21
Feature/remove cost of change

yancyribbens:
Remove cost_of_change param.   Use InputWeightPrediction to compute the cost of creating the change output instead.


yancyribbens:
stale

PR #58, opened 2024-07-17
Add WeightedUtxo trait and trait bound in place of struct

yancyribbens:
Add a new trait `WeightedUtxo` and use that trait as a trait bound for select coin function calls.  This replaces the previous concrete type WeightedUtxo.

```
pub fn select_coins<Utxo: WeightedUtxo>(
    target: Amount,
    cost_of_change: Amount,
    fee_rate: FeeRate,
    long_term_fee_rate: FeeRate,
    weighted_utxos: &[Utxo],
) -> Option<impl Iterator<Item = &Utxo>> {
```


yancyribbens:
I don't expect any review on this, and there is no change to the core algorithm competency.  Will merge and publish a new version after leaving open a short while.

yancyribbens:
Added a commit to protect SRD from panic so it can go out when the next release is published.  Also added a minor correction to the README.

PR #60, opened 2024-07-27
Add BnB proptest

yancyribbens:
Add a prop test for BnB

1) Create a random pool of coins that is at most 10 coins.  Limit to 10 so that the iteration limit of 100,000 isn't exceeded.
2) Add up all the even valued coins and use that as the target (hypothetical spend amount).
3) Assert that BnB finds a sum of coins from the pool which is the sum of the evens

run `cargo test` in `./prop` directory.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
 members = [
-  "bitcoin-coin-selection", "fuzz",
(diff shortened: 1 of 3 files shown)


yancyribbens:
Added arbitrary_target test.

yancyribbens:
Switching to work on arb_test

PR #63, opened 2024-09-14
Add arbtest

yancyribbens:
Adds proptests using arbtest to test the correctness of the solutions provided by the SRD and BnB search algos.  I use the dep exhaustigen to build all possible solutions and then run SRD or BnB and compare the results to show that the correct solution is found.  Also, BnB contains a few proptest where a UTXO pool is randonly generated, and either one or many of the UTXOs are selected as a target with the assertion that BnB will find those UTXOS as the solution while searching.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,14 +14,27 @@ keywords = ["bitcoin", "coin-selection", "coin", "coinselection", "utxo"]
 readme = "README.md"
 
 [dependencies]
-bitcoin = "0.32.3"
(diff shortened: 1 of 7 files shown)


yancyribbens:
Set maximum pool size to 20 since I had a problem where arbitrary pool size caused cargo to time out after 60 secounds

yancyribbens:
I have been running the proptests locally for the last 150 days without error.  Merging.

PR #67, opened 2024-10-29
Apply previous UTXO shortcut optimization

yancyribbens:
If the index_selection is not empty, and if the previous UTXO was excluded, then check the effective_value of the current UTXO.  If it has the same effective_value then skip searching this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ A basic performance comparison between this Rust BnB implementation and [Bitcoin
 
 |implementation|pool size|ns/iter|
 |-------------:|---------|-------|
-|      Rust BnB|    1,000|897,810|
+|      Rust BnB|    1,000|598,370|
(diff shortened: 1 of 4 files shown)


yancyribbens:
rebase

yancyribbens:
rebase, no change other then a format commit.

yancyribbens:
This PR has been open for nearly 6 months now without feedback.  I've been running continues property tests without finding any mistake.  Since this does change the core algorithm, I've been holding off on updating it, although after extensive review, I'm as confident as can be about this change.  The change mirrors the Core implementation in every way except it does not compare the fee value between UTXOs.  I've started an issue to track about that parameter [here](https://github.com/bitcoin/bitcoin/issues/32047#issuecomment-2729873842) since I believe it to be at best a useless condition so I am not going to mirror that portion.

PR #69, opened 2024-12-14
Add coin-grinder

yancyribbens:
Add Coin Grinder, A Rust DFS-based selection algorithm which optimizes for transaction
weight and creates a change output.

This is a Rust implementation of the C++ version https://github.com/bitcoin/bitcoin/pull/27877

I plan to follow up and add fuzz tests, property tests and benchmarks before a release.  While i've made a best attempt to guard against possible overflows by using checked arithmetic, I suspect more will be uncovered once fuzz tests are added.  Every instance of checked arithmetic slows the performance, therefore I've left places unchecked for now that are not obvious.

### DIFF
--- a/src/single_random_draw.rs
+++ b/src/single_random_draw.rs
@@ -92,7 +92,6 @@ mod tests {
     use crate::WeightedUtxo;
 
     const FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(10);
-    const SATISFACTION_WEIGHT: Weight = Weight::from_wu(204);
 
     #[derive(Debug)]
     pub struct ParamsStr<'a> {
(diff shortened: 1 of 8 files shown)


src/single_random_draw.rs
```diff
@@ -56,7 +56,7 @@ pub fn select_coins_srd<'a, R: rand::Rng + ?Sized, Utxo: WeightedUtxo>(
 
     result.clear();
 
-    let threshold = target + CHANGE_LOWER;
+    let threshold = target.checked_add(CHANGE_LOWER)?;
```
yancyribbens, line 59: With the next release of rust-bitcoin, Amount is limited to 21 million sats, so adding `CHANGE_LOWER` to Amount::MAX can't panic, so this commit can be dropped.

PR #72, opened 2025-01-08
Update dependency commit hash

yancyribbens:
Fuzzing is currently broken as documented because the dependency commit hash is out of sync.


yancyribbens:
No logic change, only updates to fix fuzzing.

PR #76, opened 2025-03-09
Add iteration count

yancyribbens:
When refactoring or applying changes, testing the iteration count helps to catch inadvertent changes to the core algorithm competency.  Also, coin-grinder implements an iteration count, so this helps to unify the interface between the various selection algorithms.

A side note on this implementation is that if the algorithm returns None for results, then no iteration count is returned.  I'd like to
change this however it would make sense to change this when I remove the None type return and replace it with more verbose error messaging.  For now, in all tests that expect a None return, the test asserts the iteration count is zero.

### DIFF
--- a/benches/coin_selection.rs
+++ b/benches/coin_selection.rs
@@ -34,19 +34,20 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 
     c.bench_function("bnb 1000", |b| {
         b.iter(|| {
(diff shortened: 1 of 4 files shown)


yancyribbens:
This is a breaking API change that will require a version bump when released.  However, no behavior change to the core algorithm, only reveal how many iterations where taken.  Also no change to the meaning of the tests, only extend the current behavior by showing the iteration count.

PR #78, opened 2025-03-14
Update clippy to be the same as MSRV

yancyribbens:


yancyribbens:
Needed to fix a clippy warning.  No changes to the test framework logic nor core algorithms.

PR #79, opened 2025-03-14
Add utxo to string from string

yancyribbens:
* Implements  FromStr for Utxo.
* Replace helper methods with FromStr trait
* Refactor Utxo using the builder pattern.
* Add constructors to UtxoPool


yancyribbens:
No changes to core algorithms or tests, just cleanup to the test-suit to make it nicer to read.

PR #80, opened 2025-03-17
Return vector instead of iterator

yancyribbens:
An earlier implementation called collect() before returning the results. Instead of calling collect(), the API was changed to just return the iterator since it's possible dependant applications may prefer to use the iterator.  However, after a refactor, the call to collect() was removed, and now it no longer makes sense to return an iterator. Furthermore, return an opaque type makes for a more complicated return type, making it hard to alias for example.  Returning a concrete type therefore simplifies the API.


yancyribbens:
This is a breaking API change, however no changes to algorithm behavior.  Tests are not modified either except being strengthened by an iteration count assertion.

PR #82, opened 2025-03-18
Improve Docs

yancyribbens:
* Update docs so that the first sentence clearly displays when running `cargo doc --open`
* Update documentation about the return type.  It should read Vec now instead of Iter
* Add long_term_fee_rate param to docs for BnB to docs
* Add Params and Returns section to `select_coins`


yancyribbens:
No code changes, only documentation improvements.

PR #85, opened 2025-03-26
bug: test: Add panic when no results found

yancyribbens:
If the test framework asserts that a result can be found, and yet there isn't a result found, the tests should panic.  This subtle bug in the test framework was introduced in a recent commit:

https://github.com/yancyribbens/rust-bitcoin-coin-selection/commit/780d14e95c55cf785093107480e94e2f8adcc471


yancyribbens:
Fix to the test-suit, no other changes.

PR #86, opened 2025-03-28
refactor: Call helper function

yancyribbens:
`assert_coin_select()` uses default settings minimizing the number of arguments needed.  `assert_coin_select_params()` takes a struct of parameters which is more customizable.  However, both asserts are independently calling the search function, building a utxo pool, parsing arguments and asserting results.  By calling the more robust `assert_coin_select_params()` from within `assert_coin_select()`, the line count is reduced by consolidating duplicate functionality.

Remove the `FEE_RATE` const since now the fee_rate for the test is used in one place when creating the params to be passed to `asswert_coin_select_params()`.

Remove `build_pool()` for the same reason as removing `FEE_RATE` const since this is defined in the params now.

### DIFF
--- a/src/branch_and_bound.rs
+++ b/src/branch_and_bound.rs
@@ -345,25 +345,19 @@ mod tests {
         weighted_utxos: Vec<&'a str>,
     }
 
-    fn build_pool() -> UtxoPool {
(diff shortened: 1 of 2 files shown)


yancyribbens:
simple refactor to the test module, no changes to the core algorithm competency.

PR #90, opened 2025-04-04
test: Add ability to choose and assert by effective_value

yancyribbens:
A set of UTXOs for testing can be defined by either effective_value using `e()` or absolute value as default.

For example:
```rust
weighted_utxos: &["1 sat/68vb", "e(1 sat)/68 vb", "e(1 sat)/204 wu"]
```

This will evaluate to two `UTXOs` with equivalent effective_values of 1 sat even though they have different sizes (`68 vb` and `204 wu`).  Also, there will be a `UTXO` with negative effective value since the absolute value is given as `1 sat/68 vb` given some positive `fee_rate`.

Then, assert the smallest UTXO with a positive effective_value is returned:

```rust
expected_utxos: &["e(1 sat)/204 wu"]
```

Motivation:
During the selection process, input values are converted to effective_values, and then the search begins using the calculated effective_values.  In the test-suit, it's often the case that it's desired to test what happens given a particular effective_value.
Instead of manually figuring out what absolute_value is needed to result in the search routines effective_value, add the ability to denote by effective_value `e()`.  This is done by wrapping the value with e() otherwise it's assumed to be an absolute value.

Motivated by: https://github.com/bitcoin/bitcoin/pull/29532

Unlike bitcoin core's 29532 however, this PR allows for either absolute or effective_values to be defined instead of using all effective_values.  That's because there are tests that depend on the `absolute_value`, such as how the code behaves when an absolute value of `Amount::MAX` is passed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,9 +211,38 @@ mod tests {
         pub utxos: Vec<Utxo>,
     }
 
+    // TODO check about adding this to rust-bitcoins from_str for Weight
+    fn parse_weight(weight: &str) -> Weight {
+        let size_parts: Vec<_> = weight.split(" ").collect();
+        let size_int = size_parts[0].parse::<u64>().unwrap();
+        match size_parts[1] {
+            "wu" => Weight::from_wu(size_int),
+            "vb" => Weight::from_vb(size_int).unwrap(),
+            _ => panic!("only support wu or vb sizes"),
+        }
+    }
+
     impl UtxoPool {
-        pub fn new(list: &[&str]) -> UtxoPool {
-            let utxos: Vec<Utxo> = list.iter().map(|s| Utxo::from_str(s).unwrap()).collect();
+        pub fn new(utxos: &[&str], fee_rate: FeeRate) -> UtxoPool {
+            let utxos: Vec<_> = utxos
+                .iter()
+                .map(|u| {
+                    let val_with_size: Vec<_> = u.split("/").collect();
+                    let weight = parse_weight(val_with_size[1]);
+                    let val = val_with_size[0];
+
+                    let abs_val = if val.starts_with("e") {
+                        let val = val.replace("e(", "").replace(")", "");
+                        let eff_value = SignedAmount::from_str(&val).unwrap();
+                        compute_absolute_value(eff_value, weight, fee_rate)
+                    } else {
+                        Amount::from_str(val).unwrap()
+                    };
+
+                    Utxo::new(abs_val, weight - BASE_WEIGHT)
+                })
+                .collect();
+
             UtxoPool { utxos }
         }
     }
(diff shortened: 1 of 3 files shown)


yancyribbens:
Changes to the test suit, no algorithm changes.

src/lib.rs
```diff
@@ -211,9 +211,38 @@ mod tests {
...
     }
 
+    // TODO check about adding this to rust-bitcoins from_str for Weight
+    fn parse_weight(weight: &str) -> Weight {
+        let size_parts: Vec<_> = weight.split(" ").collect();
+        let size_int = size_parts[0].parse::<u64>().unwrap();
+        match size_parts[1] {
+            "wu" => Weight::from_wu(size_int),
+            "vb" => Weight::from_vb(size_int).unwrap(),
+            _ => panic!("only support wu or vb sizes"),
```
murchandamus, line 221: Nit: Please feel free to ignore. "b" is the symbol for bit, and "B" is the symbol for byte, so a virtualbyte should have a capital B. (I use "vB".)

src/single_random_draw.rs
```diff
@@ -191,7 +194,7 @@ mod tests {
         TestSRD {
             target: "1.95 cBTC", // 2 cBTC - CHANGE_LOWER
             fee_rate: "10 sat/kwu",
-            weighted_utxos: &["1 cBTC/68 vb", "2 cBTC/68 vb", "1 sat/68 vb"], // 1 sat @ 204 has negative effective_value
+            weighted_utxos: &["1 cBTC/68 vb", "2 cBTC/68 vb", "e(-1 sat)/68 vb"],
```
murchandamus, line 197: Spicy idea to allow specification of negative effective feerate. Wouldn’t that cause issues if it were to cause the absolute amount of the UTXO to be calculated as negative? (I still don’t read Rust, but do you check that the absolute value is at least 0?)

murchandamus (commented):
Good idea to allow the specification of either. I might have to adopt that.

src/single_random_draw.rs
yancyribbens: @murchandamus 

> Spicy idea to allow specification of negative effective feerate.

Glad you like the idea! I think you mean negative "effective value" here..

> Wouldn’t that cause issues if it were to cause the absolute amount of the UTXO to be calculated as negative? (I still don’t read Rust, but do you check that the absolute value is at least 0?)

The test framework would panic (as we would hope) if the absolute value was to be negative because when the `absolute_value` is calculated, it returns an `Amount` type which is unsigned.  That is, `Amount` in rust-bitcoin wraps a `u64` so it's not possible to represent a negative value and fails at compile time.

```
   pub fn compute_absolute_value(
        effective_value: SignedAmount,
        weight: Weight,
        fee_rate: FeeRate,
    ) -> Amount {
...
```

The `-> Amount` denotes this function returns an `Amount` type.

src/lib.rs
yancyribbens: Oops thanks, I'll update this.

yancyribbens: update here https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/94

src/single_random_draw.rs
murchandamus: > Glad you like the idea! I think you mean negative "effective value" here..

I did indeed!

Cool, glad you are catching it.

PR #93, opened 2025-04-11
algo: Add secondary sort by satisfaction_weight

yancyribbens:
The UTXO skip optimization: 118b1c419391943e0d80d6973fad3c07bd2ed918 introduced a possible in-efficiency such that a solution with a better waste score would be skipped.  If UTXO A is skipped and UTXO B has the same effective_value, then skip UTXO B.  However, it's possible the UTXO B had a better weight (better waste score) than UTXO A therefore it may skip a potentially lower waste solution.  In order to address this, sort the candidate selections using descending effective_value as before, however add a secondary sort of descending satisfaction_weight.  Since the optimization works by comparing the current UTXO to the previous UTXO, the new sort order ensures that the previous UTXO will always have a better waste score than current, therefore if it is skipped, a better solution is not skipped.

### DIFF
--- a/src/branch_and_bound.rs
+++ b/src/branch_and_bound.rs
@@ -186,8 +186,10 @@ pub fn select_coins_bnb<Utxo: WeightedUtxo>(
         .map(|(eff_val, waste, wu)| (eff_val.to_unsigned().unwrap(), waste, wu))
         .collect();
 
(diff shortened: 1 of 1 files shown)


yancyribbens:
Merging to prepare for release.  Further discussion can be found here https://github.com/bitcoin/bitcoin/issues/32047

PR #95, opened 2025-04-14
test: Add best weight assertion

yancyribbens:
The test might fail to find in-efficiency in the algorithms waste score calculation.  That is, the result could have a higher weight than the target_selection weight and there would be no failure in the test.  This was noticed when the following commit
118b1c419391943e0d80d6973fad3c07bd2ed918 was added but never triggered a fail in the test framework.  The commit adds an assertion such that if a result is found, the found result has a weight that is as good as the known solution or better.


yancyribbens:
Minor change to test-suit.  No algorithm changes.

PR #96, opened 2025-04-25
Switch to weight

yancyribbens:
Core uses just weight in coin-grinder, and it's complicated to maintain both satisfaction_weight and weight.  Therefore, switch to just weight units for all algorithms.


yancyribbens:
This is an API change so will go in the next planned 0.8.0 release.

PR #106, opened 2025-06-12
draft: Add Error return type

yancyribbens:
Differentiate an unrecoverable arithmetic overflow from other types of results.  For example, hitting the iteration limit or other inputs where no solution or a non-optimal solution is returned is different from parameters where a search is unable to proceed.

In order to make clear the results returned, add an error module with the following possible overflow:

* AdditionOverflow
* MultiplicationOverflow
* SubtractOverflow


yancyribbens:
replaced by https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/114

PR #108, opened 2025-06-22
draft: Heapify SRD

yancyribbens:
Add Heap data structure to SRD to track the lowest Weight Utxo's.  This will be used to enforce a max_selection_weight in SRD and next BnB.  Mirror improvement in the core implantation PR https://github.com/bitcoin/bitcoin/pull/26720.


benches/srd_selection.rs
```diff
@@ -20,11 +20,12 @@ pub fn srd_benchmark(c: &mut Criterion) {
 
     let target = Amount::from_sat_u32(50_000);
     let fee_rate = FeeRate::from_sat_per_kwu(10);
+    let max_weight = Weight::MAX;
 
     c.bench_function("srd", |b| {
         b.iter(|| {
             let (iteration_count, inputs) =
-                select_coins_srd(target, fee_rate, &pool, &mut thread_rng()).unwrap();
+                select_coins_srd(target, fee_rate, max_weight, &pool, &mut thread_rng()).unwrap();
```
murchandamus, line 28: Shouldn’t hat be in the first commit?

src/branch_and_bound.rs
```diff
@@ -25,12 +27,17 @@ pub const ITERATION_LIMIT: u32 = 100_000;
 /// None is returned.  Also, if no match can be found, None is returned.  The semantics may
 /// change in the future to give more information about errors encountered.
 ///
+/// If the maximum weight is exceeded, then the least valuable inputs are removed.  In so doing,
+/// minimize the number of UTXOs included in the result.
```
murchandamus, line 31: This comment may be fine in the context of SRD, but I don’t think it accurately portrays the behavior in BnB. Based on your commit messages, solutions in which the maximum weight is exceeded will not be considered, but it doesn’t seem accurate that the count of UTXOs would be minimized.

src/lib.rs
```diff
@@ -81,12 +81,17 @@ pub trait WeightedUtxo {
...
+/// minimize the number of UTXOs included in the result.
+///
 /// # Parameters
 ///
 /// * target: Target spend `Amount`.
 /// * cost_of_change: The `Amount` needed to produce a change output.
 /// * fee_rate:  Needed to calculate the effective_value of an output.
 /// * long_term_fee_rate: Needed to estimate the future effective_value of an output.
+/// * `max_weight` - the maximum selection weight allowed.  For `TRUC` transactions, use a weight
+///   of 10,000 vB (40,000 WU) for this parameter.
```
murchandamus, line 94: Note that the limit is 40,000 WU for TRUC transactions that only use confirmed inputs, but TRUC transactions that spend unconfirmed inputs are limited to 4,000 WU by standardness rules. In case your coin selection permits unconfirmed inputs, you will need to ensure that TRUC transactions can only spend confirmed outputs or unconfirmed outputs from TRUC transactions, and that the limit is reduced when unconfirmed outputs are consumed by a transaction.

Additionally, if you allow spending of unconfirmed UTXOs, you must ensure that TRUC outputs are not spend by non-TRUC transactions.

src/branch_and_bound.rs
yancyribbens: Good catch.  I did a copy and paste and forgot to update.  I'll update this accordingly.  Thanks!

src/lib.rs
yancyribbens: Hmm maybe it's best to just leave out the TRUC portion is this doc since there's not a clear default as I first thought.  This is meant as being a brief description of the parameters although I suppose it could be expanded to include more detail of TRUC transactions if needed also.

yancyribbens:
> I’m not sure whether this comment is outdated, but for the algorithms that operate on the UTXOs sorted by effective value, [Coin Grinder uses lower weight as a tie-breaker](https://github.com/bitcoin/bitcoin/blob/f58de8749e94c4f6d5ab4a13d098ccfc291336c7/src/wallet/coinselection.cpp#L40), and [Knapsack and BnB use lower waste as a tie-breaker](https://github.com/bitcoin/bitcoin/blob/f58de8749e94c4f6d5ab4a13d098ccfc291336c7/src/wallet/coinselection.cpp#L28).

@murchandamus No, it's not outdated, thanks for following-up.  

My comment was about secondary-sort for SRD.  It would make sense to use weight as a secondary sort, since otherwise, it's possible that a solution wouldn't be found.  Consider a case where 3 UTXOs have the same effective_value, and two of the three are required to reach the target.  If one of the three is large, such that when combined with any other it exceeds the max_weight param, then removing the lightest weight from the heap causes no solution to be found, never reaching the target.  Therefore, if the heap is a max heap by weight as a secondary sort (when effective values are tied) then a solution would be found.

I see that BnB and coin-grinder use different secondary sorts.  Darn, it would be so much more elegant to code if the same sort could be used for all selection algorithms :)

PR #117, opened 2025-07-11
refactor: create a new result vector

yancyribbens:


yancyribbens:
It would be interesting to add some simple benchmarks to SRD for changes like this.

PR #129, opened 2025-07-25
refactor: test: move target generation inside arbitrary

yancyribbens:
Move complexity of solution generation for the UtxoPool into pool creation.  Therefore multiple pools do not need to be built and combined later.


yancyribbens:
replaced by https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/130

PR #139, opened 2025-08-16
test: update to work against 0.32.x

yancyribbens:
rust-bitcoin 0.32.x is missing some components for the tests to work for coin-selection 0.7.x.  This commit uses a fork of 0.32.x that adds the needed Arbitrary types to units as well as an additional method `checked_div_by_weight` to Amount.  Renamed the function to not include `_floor` since I did not include `ceil` as well in the fork.

Lastly, rust-bitcoin `0.32.x` does not recognize `0` as a valid Amount without adding `sats`.  This was later fixed, however since the goal if this is to enable 0.7.x tests to work with 0.32.x rust-bitcoin, this is updated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,11 +14,11 @@ readme = "README.md"
 rust-version = "1.63.0"
 
 [dependencies]
-bitcoin = "0.32.5"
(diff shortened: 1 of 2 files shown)


yancyribbens:
see https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/140

PR #179, opened 2025-09-28
ancillary: impl display for WeightedUtxo

yancyribbens:


yancyribbens:
Debug trait is fine for now

PR #189, opened 2025-10-10
dependencies: specify minor version and let patch version float

yancyribbens:
There is no reason to dictate the patch version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,14 +17,14 @@ rust-version = "1.63.0"
 features = ["rand"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -17,14 +17,14 @@ rust-version = "1.63.0"
 features = ["rand"]
 
 [dependencies]
-bitcoin = "0.32.7"
```
yancyribbens, line 20: Arbitrary is on 0.32.8 only so the patch version makes sense to keep at 0.32.8

PR #258, opened 2026-07-14
Backport pr 96

yancyribbens:


yancyribbens:
I don't think back-porting the changes into 0.x.x is the most palatable way to go since the divergence between 1.0 types and 0.32.x types happens very far back.

PR #261, opened 2026-07-18
0717 create algo routines

yancyribbens:
// TODO create the concrete type `weighted_utxos` from the Trait to be introduced.
    // I'm leaning toward calling the trait Selectable or maybe Spendable.  This was my
    // thinking after observing on PR 74 of silent-payment that impl WeithedUtxo on a struct
    // called Spendable was sort of weird sounding.  impl Spendable for SpendableCoio<'>
    // would sound better.


src/branch_and_bound.rs
```diff
@@ -142,53 +142,92 @@ pub const ITERATION_LIMIT: u32 = 100_000;
...
     weighted_utxos.sort_by(|a, b| {
         b.effective_value().cmp(&a.effective_value()).then(a.waste().cmp(&b.waste()))
     });
 
-    if available_value < target {
-        return Err(InsufficentFunds);
+    let result = bnb_select(available_value, target, bound, max_weight, &weighted_utxos);
+    match result {
+        Ok((iters, selected, weight_exceeded)) => {
+            let result = selected.into_iter().map(|i| &spendable_coins[i]).collect();
```
yancyribbens: Instead of using collect() to allocate a vector that's returned, it would be better to return `impl Iterator<Item = &T>`

src/branch_and_bound.rs
```diff
@@ -142,53 +142,100 @@ pub const ITERATION_LIMIT: u32 = 100_000;
 //
 // If either 1 or 2 is true, we consider the current search path no longer viable to continue.  In
 // such a case, backtrack to start a new search path.
-pub fn branch_and_bound<'a, T: IntoIterator<Item = &'a WeightedUtxo> + std::marker::Copy>(
+pub fn branch_and_bound<T: Spendable>(
```
yancyribbens, line 145: change to `T: Spendable + IntoIterator`. them parameter `spendable_coins: &[T],` becomes just `T`.

yancyribbens:
replace with https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/276

PR #267, opened 2026-07-24
set min weight for WeightedUtxo

yancyribbens:


src/weighted_utxo.rs
yancyribbens, line 92: positive_effective_value should be greater than zero.

PR #288, opened 2026-08-06
0806 remove checked arithmetic for waste

yancyribbens:
Checked arithmetic should no longer be needed with the new Amount bounds.


yancyribbens:
superseded by: Hang in there while we check the branch's status.

PR #333, opened 2026-08-31
refactor: return overflow as error

yancyribbens:
With the new bounds on Amount, remove superfluous check for overflow.


yancyribbens:
Ran fuzz test for 24 hours without failure.

PR #334, opened 2026-08-31
Saturate to Amont::MAX for cost_of_change

yancyribbens:
The search can still succeed even if the calculating the upper bound overflows by simply saturating the upper bound to Amount::MAX.  Furthermore, logic is simplified by not needing to handle an overflow error.


yancyribbens:
successfully fuzz tested for 24 hours

yancyribbens:
These two commits should have been squashed before merge.  I'm not going to bother dirtying the commit history anymore to try and correct that though.

